PR #32, opened 2018-03-08
Add a double backslash for windows paths to load sketches

jlchapa:
Tried to use the project but got an exception in the eval("require.... " looks like paths in windows need double backslashes


funwithtriangles:
Awesome, thanks! :)

PR #43, opened 2018-03-23
Added audio shaping sliders

cale-bradbury:
show/hide the sliders by clicking the spectrum preview

normalize slider is useful when working with a low volume feed, stores the maximum value it has received for each bin and adjusts the output acordingly to fill the full 0-1 range

falloff slider creates a slow release after the level peaks above the current value, slider is the amount it reduces each frame (value of 1 reduces fully, effectively turning it off)

### DIFF
--- a/src/components/AudioAnalyzer/index.js
+++ b/src/components/AudioAnalyzer/index.js
@@ -27,16 +27,15 @@ class AudioAnalyzer extends React.Component {
   componentWillUnmount () {
     uiEventEmitter.removeListener('slow-tick', this.draw)
   }
-  
+
   handleMouseDown (e) {
-    this.isOpen = !this.isOpen;
+    this.isOpen = !this.isOpen
     this.forceUpdate()
   }
 
   draw () {
     inputs = this.context.store.getState().inputs
     bands = [inputs.audio_0.value, inputs.audio_1.value, inputs.audio_2.value, inputs.audio_3.value]
-    
     this.drawGraph(bands)
   }
 
@@ -51,7 +50,7 @@ class AudioAnalyzer extends React.Component {
       val = data[ i ]
       height = this.height * val
       offset = this.height - height - 1
-      hue = i / data.length * 360  
+      hue = i / data.length * 360
 
       this.ctx.fillStyle = 'hsla(' + hue + ', 100%, 50%, 1)'
       this.ctx.fillRect(i * this.barWidth, offset, this.barWidth, height)
@@ -63,13 +62,13 @@ class AudioAnalyzer extends React.Component {
   render () {
     return (
       <div>
-      <canvas ref={node => { this.canvas = node }} 
-      onMouseDown={this.props.onMouseDown || this.handleMouseDown}
+        <canvas ref={node => { this.canvas = node }}
+          onMouseDown={this.props.onMouseDown || this.handleMouseDown}
       />
-      {this.isOpen &&
-        <div style={{position:"fixed"}}> 
-          <Control nodeId="audioNormalizeLevels" />
-          <Control nodeId="audioLevelsFalloff" />
+        {this.isOpen &&
+        <div style={{ position:'fixed' }}>
+          <Control nodeId='audioNormalizeLevels' />
+          <Control nodeId='audioLevelsFalloff' />
         </div>
       }
       </div>
@@ -79,6 +78,9 @@ class AudioAnalyzer extends React.Component {
 
 AudioAnalyzer.contextTypes = {
   store: PropTypes.object.isRequired
+}
+AudioAnalyzer.propTypes = {
+  onMouseDown: PropTypes.func
 }
 
 export default AudioAnalyzer
--- a/src/components/AudioAnalyzer/index.js
+++ b/src/components/AudioAnalyzer/index.js
@@ -1,6 +1,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import uiEventEmitter from '../../utils/uiEventEmitter'
+import Control from '../../containers/Control'
 
 let height, val, offset, hue, i, inputs, bands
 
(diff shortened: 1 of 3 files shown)


funwithtriangles:
Awesome! I'll have a look through this tonight.

funwithtriangles:
I see some linting errors, do you have linting set up in your editor?

cale-bradbury:
haha, nope, I'll give that a try later tn :)

funwithtriangles:
Look for "eslint" plugins for whatever editor. It should detect the style rules and show you what has gone wrong. Many editors have an option to "fix on save" for convenience.  I got into the linting game late, absolutely love it now!

Also, this should fix a bunch of linting errors straight away: `yarn lint:fix`

cale-bradbury:
cool, seems all good now (at least in terms of formatting) thanks for the lint tips!

src/components/AudioAnalyzer/index.js
```diff
@@ -54,13 +61,26 @@ class AudioAnalyzer extends React.Component {
 
   render () {
     return (
-      <canvas ref={node => { this.canvas = node }} />
+      <div>
+        <canvas ref={node => { this.canvas = node }}
+          onMouseDown={this.props.onMouseDown || this.handleMouseDown}
+      />
+        {this.isOpen &&
+        <div style={{ position:'fixed' }}>
```
funwithtriangles: We're using [`styled-components`](https://github.com/styled-components/styled-components) throughout the app, instead of inline styles. To give you an idea, it would look like this:

```javascript
import styled from 'styled-components'
import theme from '../../utils/theme'
...
const SettingsBox = styled.div`
   background: ${theme.bgColorDark1}; // use the theme to get a bg color
   position: absolute; // absolute makes more sense than fixed
`
...
<SettingsBox>
   <Control nodeId='audioNormalizeLevels' />
   <Control nodeId='audioLevelsFalloff' />
</SettingsBox>
```

src/components/AudioAnalyzer/index.js
```diff
@@ -25,10 +28,14 @@ class AudioAnalyzer extends React.Component {
     uiEventEmitter.removeListener('slow-tick', this.draw)
   }
 
+  handleMouseDown (e) {
+    this.isOpen = !this.isOpen
```
funwithtriangles, line 39: While it definitely works having `this.isOpen`, all other parts of the app work by taking from the `ui` part of the state. By keeping all state at the top, it means that this little popup can be controlled by other parts of the app. For instance, we could fire an action when the body is clicked to hide it.

In order to do this, I would:

- Create a "container" for AudioAnalyser. This is a redux thing and is a "higher order component" that makes use of `mapStateToProps` and `mapDispatchToProps`. By doing this, you can pass `isOpen` and `onClick` to the component as props.
- To actually make changes to the state, have a look in `store/ui`. Currently there is an `isEditing` property. Up until now it has been used to determine whether popups for editing node titles are showing or hiding. However there's no reason not to use it for this. Something like:

```javascript
onClick: dispatch(uiEditingOpen('audioAnalysisPanel')
```

Then in your container you'd be able to do something like:

```javascript
isOpen: state.ui.isEditing && state.ui.isEditing.type === 'audioAnalysisPanel'
```

If you want to be super neat, write a selector so that it just looks like:
```javascript
isOpen: getUiIsEditingType(state) === 'audioAnalysisPanel'
```

funwithtriangles: You'd probably actually want to make a new action called `uiIsEditingToggle()` and add a new section to the reducer to handle it. Something like:

```javascript
case 'UI_EDITING_TOGGLE': {
      return {
        ...state,
        isEditing: state.isEditing 
          ? false 
          : {
            id: p.id,
            type: p.type
          }
      }
    }
```

src/inputs/AudioInput.js
```diff
@@ -1,13 +1,32 @@
...
+      type: 'param',
+      key: 'normalizeLevels',
+      value: input.normalizeLevels,
+      id: 'audioNormalizeLevels'
+    }))
+
+    store.dispatch(uNodeCreate('audioLevelsFalloff', {
+      title: 'Levels Falloff',
+      type: 'param',
+      key: 'levelsFalloff',
```
funwithtriangles: I'm not sure if there's any need for `key` here but there's no harm in it being here. The `key` is used for params.

funwithtriangles (commented):
I've pulled this down it works great! Awesome stuff, thanks. Below are some comments to explain what I'd have done differently. I'm happy to make the changes but I know you might like to have a go. :)

src/components/AudioAnalyzer/index.js
cale-bradbury: Got the other two comments taken care of, but would you be able to get this last one? Poked around a bit and it's an amount over my head at the moment. Don't know when I'll have the time to trial and error my way into creating a container. I'll definetly be reading your commit though to get a grasp on how it's done in the future.

funwithtriangles:
Cheers Cale! I've merged it into `feature/audio-shaping-sliders` so you can see the work I do on there before merging into `dev`. :)

funwithtriangles:
That's now already merged into `dev`! Take a look at the commits on that branch to see the extra things I did. 

Container/reducer/action:
https://github.com/nudibranchrecords/hedron/commit/92a93da3c7775e8cd9a9b9b6610d670739123a5d
https://github.com/nudibranchrecords/hedron/commit/050643cd42b94f00de2df34c0784b6fc5c76de72

Also worth noting that I've moved those initial dispatches of `uNodeCreate` into a new place. This is to get it inline with some upcoming changes involving scene switching,
https://github.com/nudibranchrecords/hedron/commit/24971fa1dd5b30f465d4c6ea600fe642def80ff1

Let me know if you have any questions!

cale-bradbury:
Thank you so much <3 I'll dive in to this when I'm able to hack some more in a week or two! Going to get in an audio data texture similar to shadertoys to pass into THREEjs/shaders next!

funwithtriangles:
Wow, sounds exciting! Looking forward to it.

funwithtriangles:
Perhaps worth creating an issue to explain and discuss the approach?

Issue #13, opened 2018-02-22
Param improvements (types and groups)


funwithtriangles:
Currently, params are only ever a value between 0 and 1. It would be useful to be able to have some extra types such as boolean params, option params (selectable distinct values), 3 part vector params (colour picker, positions, etc).

To get here, it might be useful to first implement param groups, which would be a way to organise params into groups that can be opened and closed. This could help with 3 part vector params.

funwithtriangles:
Exponential / Logarithmic types for params would also be useful. Should be user defined but have a default option in the config.

funwithtriangles:
Strings and enums, 3D model upload, image upload.

PR #62, opened 2018-05-10
Added paramaters to config.js parameter

cale-bradbury:
added optional `min`/`max` values to a config.js param

made `title` optional, creates one based on the `key` if not found


cale-bradbury:
Seems to not like the min/max being undefined in a param. I'm guessing the way I checked to see if the value is null or not doesn't worked with the ci

funwithtriangles:
`title` being optional is good! I also think `defaultValue` should be optional too, as I think that causes errors if omitted (should just be set to 0 if no `defaultValue` is given). 

I'm curious as to know what your reasons for wanting a `min` and `max`. Currently, the way Hedron works is that the params that come into a sketch are _always_ a number between 0 and 1. I quite like the simplicity of this. The way I work is I put any min/max stuff or any conversions of the 0-1 inside the sketch, as often these values are quite specific to what's going on inside the sketch file (distances, scales, colours, etc) and so it feels like a good place to do it. However as always I'm totally open to hear your reasoning!

Also, it's worth noting that there are already "modifiers", that allow the user to set a lower and upper range for a param.

The tests failing are just because there are now extra things appearing for the param that it doesn't expect, I can fix that up.

cale-bradbury:
Maybe doing it in the sketch is good, probably much handier for folks who are reading a sketch over.

My thoughts were that the upper/lower ranges might want to get extended while your in the middle of a show or a situation where you don't want the app to reload. The ui could also reflect the proper range, which would make it easier to tell what the 'true' value of a number is. (not that I implemented either of these two yet)

I might be confusing the modifiers you are talking about with something else, but the modifiers currently only let you change the values to numbers between 0-1, you can't extend something past 1 right?

funwithtriangles:
So you'd have an extra min/max UI slider for each param? I do actually see the need for something like this. Sometimes you have a range you want to stick with but then extend later on in a show. To me this is more of a "range" thing than a "min/max" thing. And then by setting the range the UI would then say something like 7.1 instead of 0.71? 

You're right about modifiers, they keep everything inside a range of 0 and 1. In the past I've had the lower and upper range set at something like 0.3 and 0.6 and then extended them out to mimic what you're trying to do. But it was very confusing and I stopped working that way!

As a side note, I do intend to have new types of params at one point, where it will display -1 to 1, ON and OFF, colour info, etc (https://github.com/nudibranchrecords/hedron/issues/13). However, I still think behind the scenes these values could remain as 0-1. ;)

cale-bradbury:
Yeah, I think that a solution that would let you define the range would be good, and also take care of the -1 to 1 case, perhaps getting the default range from the config.js, but then letting that be edited/saved/reverted in the main program

cale-bradbury:
removing this pr to add a new one from a better branch ( https://github.com/netgrind/hedron/tree/feature/min-max )

currently fixing the ui that displays the min/max in its popup, ought to be ready soon

PR #63, opened 2018-05-16
Fix to support a true fullscreen on MS Windows

esnho:
As announced via mail this is a little fix to support a true fullscreen window in MS Window.
User can acces to the menu bar using Alt.
Fullscreen can be toggled using F11.


funwithtriangles:
Looks good! I'll just need to test if it breaks anything in OSX / Linux. Thanks so much for contributing! :)

funwithtriangles:
Quick check on full screen modes in Linux / OSX and it all seems okay. Thanks again for this! :)

PR #64, opened 2018-05-23
Fix/lfo timing

cale-bradbury:
Noticed that LFOs weren't in sync with the beat counter, my guess was that this had something to do with not maintaining a perfect frame rate, or perhaps the pi/24 number was not what was needed? Regardless it is now computed based on the current division of the beat.

Also fixed/simplified up the math for the waves so they all oscillate over the same amount of time, sine waves were not in line with the others before


funwithtriangles:
This is great, thanks! That whole area has been a bit of a mess for a while, thanks so much for cleaning it up. Just merged into a separate branch to fix a little code style thing before merging into `dev`.

PR #65, opened 2018-05-24
Snap the tempo to the closest beat when hitting tap tempo instead of resetting it

cale-bradbury:
makes things look better/less jarring when you have stuff on lfos :)


funwithtriangles:
This looks very useful! Unfortunately it actually seems to be broken. `totalBeats` isn't defined and it's erroring when I hit Tap Tempo. Is it working your end? Perhaps you didn't commit a change?

cale-bradbury:
fixed it up, sorry about that, was flying by the seat of my pants most of the weekend :)

funwithtriangles:
Good stuff, thanks for this!

Issue #45, opened 2018-04-03
Move guides to codebase


funwithtriangles:
Move "user guide" and "creating sketches" from the Wiki and into something like `/docs`. Means this stuff can be properly synced with each release / feature.

funwithtriangles:
Docs for destructor here: https://github.com/netgrind/hedron/wiki/Creating-Sketches

funwithtriangles:
Docs now in codebase (https://github.com/nudibranchrecords/hedron/pull/82)

PR #69, opened 2018-07-09
optional destructor function in sketch gets called upon deletion, use…

cale-bradbury:
…ful if there is anything that needs to be cleaned up

I use this for post effects, and a video player (adds a hidden video element to the dom in order to get a THREE.VideoTexture)

### DIFF
--- a/example-projects/simple/sketches/solid/index.js
+++ b/example-projects/simple/sketches/solid/index.js
@@ -121,6 +121,13 @@ class Solid {
       }
     })
   }
+
+  /** HEDRON TIP **
+    Do things here for when the sketch is deleted
+  **/
+  destructor () {
+    console.log('Solid sketch deleted!')
+  }
 }
 
 /** HEDRON TIP **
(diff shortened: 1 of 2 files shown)


funwithtriangles:
This looks good, nice idea. This is also a good example of why the docs need to be part of the source code, so changes like this can also come with some docs changes.

I'm sure it all works fine but I'm yet to test it my end, will try and find the time to do this weekend. I might write a little test for it too. :)

cale-bradbury:
hmm, I couldn't really figure out how to do a pull request on the wiki :/
check my version here though, added a bit about how the scene gets passed into the constructor too!

edit: forgot link https://github.com/netgrind/hedron/wiki/Creating-Sketches

netgrind:
I seemed to have not added the link

funwithtriangles:
Merged, docs still need updating, captured in separate issue: https://github.com/nudibranchrecords/hedron/issues/45

Issue #9, opened 2018-02-22
Extra LFO options


funwithtriangles:
- [x] phase
- [x] noise shape
- [ ] random shape

funwithtriangles:
Custom text box for code based wave function (i.e. After Effects)

funwithtriangles:
Text box has own issue now: https://github.com/nudibranchrecords/hedron/issues/115

esnho:
Hi @funwithtriangles !

I'm improving the Noise function, I'd like to make a pull request but sincerely I cannot understand why a test fails.

I'm not used to these kind of tests so would be helpful if you help me.

I've basically changed the firm of the function that calls the LFO in order to pass the link id, this is useful with the improved OneDimensionalNoise.
Finally the returns are link based and can be used to obtain different noise oscillators around the scene ;)

Please, can you help me with the tests?

Here is the fork link: https://github.com/esnho/hedron

This is what test returns:

```
(Saga) handleInput (lfo)
    √  0. Call debounceInput
    √  1. Gets assigned links
    √  2 Get Options values (nodes)
    ×  3. get value after going through first modifier at Test.<anonymous> (D:/_p/milano-live-coding/hedron/src/store/inputs/_test/sagas.spec.js:340:5)
        { '@@redux-saga/IO': true, CALL: { context: null, fn: [Function: _default], args: [ 0.555, 'sine', 2, 0.25, undefined'id' ] } }
{ message: 'Error!' }
    √  4. Dispatches batch node update action
    √  Dispatches project error if some error
```

funwithtriangles:
Hi @esnho. Thanks for this! Feel free to make the pull request and I can take a look, the tests are a bit of a mess at the moment and I'm thinking about disabling most of them for the time being, so don't feel disheartened about them failing! When you make the PR just make sure I have the ability to commit to the branch. :)

esnho:
Hi @funwithtriangles ! Here's the pull, to be sure you can edit the source I've added you as collaborator :)
https://github.com/nudibranchrecords/hedron/pull/145

PR #70, opened 2018-07-17
added cosine lfo

cale-bradbury:
Very important to have both sine and cosine to create good 3d wobbles. In the future, we can add a phase parameter, but not quite sure at the moment on how to work that data in with the current modifier setup.

### DIFF
--- a/src/utils/lfoGenerateOptions/index.js
+++ b/src/utils/lfoGenerateOptions/index.js
@@ -16,6 +16,10 @@ export default () => {
           label: 'Sine'
         },
         {
+          value: 'cosine',
+          label: 'Cosine'
+        },
+        {
           value: 'square',
           label: 'Square'
         },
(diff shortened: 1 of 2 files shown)


funwithtriangles:
Thanks for this! Phase is planned (https://github.com/nudibranchrecords/hedron/issues/9) but I'd want to do it properly with divisions of the beat, etc. This looks like a nice easy interim! 

I guess the only little issue I can foresee is that once phase is implemented, we may want to remove cosine and this could affect old project files.

cale-bradbury:
good point! could have it default to sine since that is at the top of the list if a wave is not found, but yeah, maybe more trouble than it's worth, feel free to close this if so :)

funwithtriangles:
Closing in favour of https://github.com/nudibranchrecords/hedron/issues/9

PR #71, opened 2018-07-17
Feature/param return

cale-bradbury:
lets a shot return an object containing an updated set of parameters, handy if you want to store the state of what a function call does, make a call that randomizes variables, make a call that sets things to a specific state

### DIFF
--- a/example-projects/simple/sketches/solid/config.js
+++ b/example-projects/simple/sketches/solid/config.js
@@ -28,6 +28,11 @@ module.exports = {
       key: 'scale',
       title: 'Scale',
       defaultValue: 0.5
+    },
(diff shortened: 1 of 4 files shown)


cale-bradbury:
might be handy to create a `hidden: true` flag in the param config in the future as well (perhaps with an in editor 'debug mode' to let you view them)

funwithtriangles:
This is really great! I've wondered about how to do this before, nice solution. I've made a little performance change and also updated the example project.

funwithtriangles:
Hidden params captured as a separate issue: https://github.com/nudibranchrecords/hedron/issues/78

PR #74, opened 2018-07-30
Improves normalized values, adds more audio shaping options

cale-bradbury:
Normalization now tracks the lowest value it has observed in addition to the highest, this helps with signal noise, values that bounce nicely but don't drop to 0.

Normalization Range Falloff has a more logical default value that will produce better results

Values can now be smoothed over time via a simple lerp towards the raw value using a slider

Values can undergo an exponential operation to help give everything a boost, or ensure that only the highest values really make things react


cale-bradbury:
One issue with this PR is that it breaks old sketches, seems that projects don't update nicely to potential additions to `setCoreState.js`, going to look into having it automatically check for missing elements on another day, just thought I would get this PR started in case this is something you know how to fix quickly @funwithtriangles

cale-bradbury:
Also wondering if it makes sense to generate a data texture within hedrons core, there are a lot of pros and cons to both routes, I could leave this part within my personal branch for now

cale-bradbury:
Just remembered that this PR relies on #75  to be merged in first!

funwithtriangles:
What's the situation with this one? @netgrind Do you want to merge the changes from dev into this branch and see if it's all working properly?

cale-bradbury:
Seems to be working as expected :) tunes hitting the highest highs and the lowest lows

funwithtriangles:
All looks good. I noticed that when dragging the values, if you released your mouse click outside of the panel, it would hide it, which I felt was quite annoying. I've now fixed this so if you released the mouse click when outside of it, using `onMouseDown` instead of `onClick`. This is more of an app-wide change than specific to this feature (the top `App` component is always looking out for an `onClick` which I've now changed to `onMouseDown`).

Also in the same commit swapped out `onClick` props to `onMouseDown` for time sensitive actions (e.g. the tap tempo stuff)  this should probably have been done in a separate PR but oh well!

Before merging, I was just wondering if you had the time to write a tiny sentence for each value in user guide? I was intrigued as to why 'Levels Power' seems to reduce the height of the bars as the value is increased.

funwithtriangles:
Thanks for all your work on this!

cale-bradbury:
no worries, hope it treats your params well on the field

PR #75, opened 2018-07-30
added min/max range to sketch params

cale-bradbury:
Params can have an optional min and max value, defaults to 0 and 1 if not found. This allows for a sketch creator to determine what numbers they would like in the config so that the data is as they would want it when it reaches their update function. A good example is setting a params max to 360 so it can be used as a hue.

The user can also then change their min/max values in the params extended controls to fit their needs or to be able to extend a range mid-show


cale-bradbury:
Not sure why the tests are failing, seems to be implying that the object has 2 keys named 'title' or the keys got rearranged..? doesn't really make sense to me how that could be happening

funwithtriangles:
I fixed the tests. The reporting in the terminal is pretty terrible at the moment so apologies for that. It often looks like the error being reported is to do with the order of the object, but it actually isn't. 

The error was there because had `min` and `max` set for a shot in one of the tests, which was failing because it shouldn't be there. There was also one other place where `min` and `max` should have been there but weren't. :)

This feature is cool and I do understand the need for it now! Few things:
- It doesn't seem to work immediately. After creating a sketch and changing the max value, the param doesn't update it's value accordingly. You need to change the min as well to see the change. After that it all seems to work fine. I think this might be some initialisation issue with Redux Form. I can have a look. Probably not worth worrying about until the UI is complete.
- The UI could be improved. I wonder if this feature could be hidden as an extra "settings" modal box? We could put other little things in here to keep the main part of the UI clean?
- Not sure if we need a "Set" button. Might be easier just for the setting to update when the form field changes value?

cale-bradbury:
I reason I went for a set button is so it won't update and cause weird jumping when you are typing in a multi-digit number. I'll look into sending an update when focus leaves the box and adding a collapsible menu

funwithtriangles:
Hmm I can see why you wouldn't want weird jumping. However I feel this
feature is more likely to be used during creation as opposed to showtime,
or do you not think? It's a toss up between easier/faster creation and
tidier showtime.

For the collapsible menu, I think the best route to go is to label it as
something like "advanced", so we can house other features in there in
future.

funwithtriangles:
@netgrind Just letting you know that docs are now part of the source code, so feel free to add some little description there as part of this PR.

cale-bradbury:
Think this may be good to merge now @funwithtriangles :)
I'd like to get your thoughts on the new Collapsible component, maybe the container script is unneeded?

src/selectors/getSketchParams.js
```diff
@@ -8,7 +8,7 @@ const getSingle = (state, sketchId) => {
 
   sketchParams.forEach((id) => {
     const param = state.nodes[id]
-    if (param.min && param.max) {
+    if (param.min !== undefined && param.max !== undefined) {
```
funwithtriangles: Min/max wasn't working at all if one of the values was set to zero, as it being seen as false. This is a good example of why tests are useful (see above test added), especially for complicated selectors. I often write the test first in these cases.

funwithtriangles, line 11: Min/max wasn't working at all if one of the values was set to zero, as it being seen as false. This is a good example of why tests are useful (see above test added), especially for complicated selectors. I often write the test first in these cases.

src/containers/Range/index.js
```diff
@@ -17,8 +17,8 @@ const mapStateToProps = (state, ownProps) => {
 
 const mapDispatchToProps = (dispatch, ownProps) => {
   return {
-    onSubmit: (value) => {
-      dispatch(nodeRangeUpdate(ownProps.nodeId, value))
+    onSubmit: (obj) => {
+      dispatch(nodeUpdate(ownProps.nodeId, obj))
```
funwithtriangles: `nodeUpdate` is an easy way to change the properties of a node without having to create a custom action/reducer for it. Also, just note that all of your handling of this was in the macros section for some reason.

src/store/macros/sagas.js
```diff
@@ -180,16 +180,6 @@ export function* handleNodeValueUpdate (action) {
...
-export function* handleNodeRangeUpdate (action) {
-  const p = action.payload
-  const node = yield select(getNode, p.id)
-  if (p.value.min) {
-    node.min = p.value.min
-  }
-  if (p.value.max) {
-    node.max = p.value.max
-  }
-}
```
funwithtriangles: Although this code has been removed, it's worth noting that you should always dispatch an action to modify the state inside the reducer (using immutable methods) rather than directly modify the variable as you are doing here.

src/components/SketchParam/index.js
```diff
@@ -1,16 +1,26 @@
...
+  border-bottom: ${theme.lineColor1} 1px dashed;
+`
 
 const SketchParam = ({ nodeId, sketchId }) => (
   <Param nodeId={nodeId} sketchId={sketchId}>
-    <Collapsible {... { title:'Advanced' }}>
+    <Top>
+      <InputLinkUI nodeId={nodeId} />
+    </Top>
+    <Revealer title='Advanced' isOpen>
```
funwithtriangles: I generally don't like using little libraries such as the one you implemented. It's quick and easy to implement but then when you want to do something a bit more custom, things get hacky. I made a little component called `Revealer` that does the same thing. :)

src/store/ui/reducer.js
```diff
@@ -55,6 +56,21 @@ const uiReducer = (state = defaultState, action) => {
         openedNode: p.id === state.openedNode ? false : p.id
       }
     }
+    case 'UI_AUX_TOGGLE_OPEN': {
```
funwithtriangles, line 59: I created a generic action/reducer thingy that allows things to be opened and closed globally. All it does is add or remove keys from an array. If a key is present in the array, we know this thing is "open". Currently using this for the "advanced" menu reveal for params. Could be used anywhere else, as long as the keys are always unique.

src/containers/SketchParam/index.js
```diff
@@ -11,6 +12,9 @@ const mapStateToProps = (state, ownProps) => ({
 const mapDispatchToProps = (dispatch, ownProps) => ({
   onAdvancedClick: () => {
     dispatch(uiAuxToggleOpen(ownProps.nodeId))
+    setTimeout(() => {
+      uiEventEmitter.emit('recalc-param-heights')
```
funwithtriangles, line 16: The way the param boxes are styled is a bit strange and when they open/close and can cause overlapping of elements when heights change. The UI needs to be told that heights have changed. The way I've been handling this is with a little event emitter.

funwithtriangles:
@netgrind I've left a bunch of comments (you can see them in the activity log above). Let me know if you have any questions on what/why I've changed stuff. I'd recommend pulling down and having a look. :)

Leaving open for now to give you a chance to review what I've done.

funwithtriangles:
Also, do you think it makes more semantic sense in the config for it to be `defaultMin` and `defaultMax`?

src/selectors/getSketchParams.js
cale-bradbury: Ahh, I'll be more verbose in the future!

src/store/macros/sagas.js
cale-bradbury: I take it that that is to help with undo history or something similar?

cale-bradbury:
Thanks so much for the comments and code, this almost all makes sense. I'm a bit confused at the component not opening/closing itself when it's clicked, I assume it's because you want the action to have been dispatched and passed through the store?

cale-bradbury:
One thing I noticed that still remains, with regards to the param bar, do we want that to display the min/max, it's kind of a toss-up as sometimes it's easier to type in an absolute value, while others it is easier to input a ratio, perhaps pick one, but then we could add a toggle in the future?

funwithtriangles:
> Thanks so much for the comments and code, this almost all makes sense. I'm a bit confused at the component not opening/closing itself when it's clicked, I assume it's because you want the action to have been dispatched and passed through the store?

I'm trying to follow the design principle of using "Stateless Functional Components" where possible. This keeps the components as "stupid" as possible, so all app logic sits outside of these presentational parts of the application. There are [many benefits](https://hackernoon.com/react-stateless-functional-components-nine-wins-you-might-have-overlooked-997b0d933dbc) to doing it this way, for me the main one being that we can control any component from outside of the component. For instance, let's say we needed to close the "advanced" panel programatically from somewhere else in the app. This can't be done when the state only lives inside that component. If we're keeping the state for that component up the top in the redux store, we can easily do it with an action.

> One thing I noticed that still remains, with regards to the param bar, do we want that to display the min/max, it's kind of a toss-up as sometimes it's easier to type in an absolute value, while others it is easier to input a ratio, perhaps pick one, but then we could add a toggle in the future?

I'm leaning towards keeping it as a value between 0 and 1, mainly for presentation reasons. It's more predictable to know that it's only ever going to be 4 digits long (e.g 0.123). If someone had a min of (-1000000) and a max of (1000000) it might mess up the layout. However I do see the reason why you might want to type in an absolute value in certain cases. For now I'd prefer it to stay at 0 and 1 but we can revisit later?

Also, do you think it makes more semantic sense in the config for it to be `defaultMin` and `defaultMax`?  It's a number that can be changed in the interface, just like `defaultValue`. See what I mean?

cale-bradbury:
yeah, defaultMin/Max makes sense, I'll work on changing that and add a
button in advanced to reset the min/max to default (currently resets on
reimport which is kind of actually very undesired imo)

cale-bradbury:
Seems to be another issue with key order (or duplicated keys) in the unit tests, min/max is behaving much better now, I added a button to reset it under Advanced. Don't have time to try and get yarn test working properly on windows atm, could you give me a ping if you could confirm it working for you/fixed @funwithtriangles ? No rush

funwithtriangles:
Sure, I can fix this up when I get the chance! Good excuse for me to look at how tests work on Windows.

funwithtriangles:
So the tests will always pick up on some genuine error but are reporting it terribly. The issue was based around the idea that if a sketch param does not yet have a min/max key associated with it, re-importing sketches will not fix this (as the tests were asserting). For now I've just added the min/max keys in the test, assuming that the sketch params already has min/max keys. I'm not too bothered about backwards compatibility at this stage.

As for the test issues, it's clear that this is a real issue going forward. I've been using tap until now to keep things minimal but I'm starting to see this is quite unsupported.  It looks like [jest](https://jestjs.io/) is the way to go. Will make a new issue for the migration over to a new testing library.

funwithtriangles:
Either way, thanks for this feature! Merged. :)

Issue #89, opened 2018-09-07
Edit subnodes with text field


funwithtriangles:
Params can be edited by double clicking and typing, this isn't possible for subnodes (e.g. modifiers).

funwithtriangles:
Fixed https://github.com/nudibranchrecords/hedron/pull/91

PR #90, opened 2018-09-07
Feature/lfo phase

funwithtriangles:
LFO phase. Use 0.25 for cosine. Made a simple little trig example project, worth checking that out to see it all in action. @netgrind  can you look over this and see if you think it all makes sense? I feel it's all good for sine/triangle/square. Not sure about the sawtooth ones?

Note: At the moment it's a bit annoying if you want to get an exact value (e.g. 0.25, 0.5), I'm fixing this separately. (https://github.com/nudibranchrecords/hedron/issues/89)


cale-bradbury (approved):
Looks great! I agree that it would be great to type the number in, but that is an issue that all those fields have, and is a separate feature to add.

funwithtriangles:
Thanks for taking a look. Already fixed the other issue, you can now double click on any slidey thing. :)

PR #96, opened 2018-09-17
Icon and Title added to the window

cale-bradbury:


funwithtriangles:
Looking great! I love the overall concept for the logo, however I think it could be tweaked slightly. For the 3D model I'd make the "arms" all a bit thicker. It feels a bit too spindly, if you get what I mean?

Also, for the actual icon, I'd probably do a traced version in Illustrator or something, so we can have a vector file and have full control over the colours.

cale-bradbury:
Cool, I'll give it a go, mostly just wanted to get the files I had pushed to git. Really like that the inner portion looks like an eye now :)

cale-bradbury:
image
image
 ended up going with the 16x16 icon, looks the best given the size, feel that maybe people might have a different icon size in their os, not sure if this can be set to something that's not a png when it isn't being compiled for a platform, been going off [this tutorial](https://www.christianengvall.se/electron-app-icons/)

funwithtriangles:
Looks awesome! I'll have a look on OSX / Linux when I get the chance.

funwithtriangles:
Here are my results with 16px, 32px and 64px icons, on Windows 10.

**Taskbar**
taskbar-16
tasbar-32
taskbar-64

**Top bar**
topbar-16
topbar-32
topbar-64

As you can see the taskbar icon at 16x16px is quite pixelated. I don't see much difference between 32 and 64 in the taskbar but it seems to look nicer at 32 for the top bar. For Windows I'd say we go for 32px. I'm not with my OSX or Ubuntu machines right now but will test them next time I am.

Worth mentioning that for full control of the top bar, we could go for a [frameless window](https://electronjs.org/docs/api/frameless-window#frameless-window).

cale-bradbury:
Yeah, really confused as to why it doesn't seem to support multiple sizes and pick the best one based on the context.

Could do a borderless window, we would need to add some controls to fullscreen/minimize/close ourself though right?

funwithtriangles:
So the tutorial you were following wasn't quite right, as we're using `electron-builder` instead of `electron-packager` (yep). All that's actually required is a couple of icon files added to a `build/` folder in the root of the project.

During dev the icon will still look a bit crap, you need to package the app for everything to look nice:
taskbar-new
topbar-new

funwithtriangles:
Looks great on OSX too!

screen shot 2018-10-16 at 19 31 35

Haven't tested on Linux but apparently it uses the same files as OSX so it _should_ be okay. Will do a proper test when it comes to deployment of 0.5.

Issue #114, opened 2018-10-07
change constructor argument orders


cale-bradbury:
I would like to propose switching the order of sketch constructor arguments from `(scene, meta, params)` to `(scene, params, meta)` since I feel params are more often needed (setting default states for shot related hidden param type things)  and meta currently only contains a single variable (that can be accessed via `__dirname` already) this is also a thing where the earlier a switch is made the less painful it will be.

funwithtriangles:
Completely agree!

funwithtriangles:
Note: Worth investigating the `__dirname` thing. I have a feeling it doesn't quite work due to the sketches being loaded dynamically but I could be wrong!

cale-bradbury:
I haven't had an issue with `__dirname` yet, but also have not stress-tested it

funwithtriangles:
FIxed https://github.com/nudibranchrecords/hedron/pull/116

PR #116, opened 2018-10-08
switched params and meta order when a sketches constructor is called

cale-bradbury:
in regards to #114 params is passed to the sketch before meta.

```🙏 rip sketches, let us complete this refactor so that others will not have to 🙏```


funwithtriangles:
Thanks for this! I'm sure it works but as a habit I still want to test it personally before I merge. Will do this soon along with some of the other PRs!

Issue #123, opened 2018-10-17
Trailing commas in lint


cale-bradbury:
If we could set the linter up to want trailing commas in objects it will help make things much more likely to merge smoothly, I find that has been a trouble point with merging on multiple occasions
```
let obj = {
    foo: 5,
    bar: 6,
}
```

funwithtriangles:
I completely agree. I love the idea of a standard set of JS lint rules
using `standard` but this is the one rule I really don't get on with. It
should be possible to override this one rule.

funwithtriangles:
Fixed https://github.com/nudibranchrecords/hedron/pull/125

PR #125, opened 2018-10-17
Trailing commas required on objects

ghost:
#123 If I understand correctly, this rule should solve your problem. I tried  this on my local machine and works well. It complete my objects with that last comma and warn an error when it is not in there.

👍 

*Edit:* I see on the travis build that now is requiring that missing comma dangle, it's this ok? that is are you waiting for?


cale-bradbury:
Those look like good errors to have 😃 try running `yarn lint --fix` in the console to have it add the commas and fix those all up

funwithtriangles:
I do love myself a bit of `comma-dangle`, thanks for this @Ranacode! A few things:

- As @netgrind has suggested, it's quite simple to fix those issues using `yarn lint --fix`
- What are your reasons for changing the spacing/formatting of this file? I feel it's less readable now
- Perhaps it might be best if we go for `comma-dangle: ["error", "always-multiline"]`?

ghost:
Oops sorry, I forgot turn off my personal format in the editor, in respect of the "always-multiline" I think it's better than just "always" :).

Sorry for the format again, see you soon :)

funwithtriangles:
We be danglin' from now on

PR #145, opened 2019-01-26
Added a better noise function

esnho:
This noise function permit to differentiate output based on link id.

There's also a new dev-dependency used to set env variable using the same command on every platform.


funwithtriangles:
Thanks for this! Will check it out soon :)

funwithtriangles:
I see what you've done here, makes sense! To make sure the noise is always different from one place to the next. I'm just wondering if there would ever be an instance where the user may want the noise to be the same. Let's say they are using it to affect the scale and transparency of something, and wanted these properties to be in sync with each other (e.g. the object fades as it grows). In this instance they'd want it that noise value to be the same.

Perhaps we can add a "seed" value? That way the user can decide whether it's going to be different or not. What do you think? Let me know if that's something you're comfortable with trying to add, otherwise I can do it. At this moment it would mean that all of the other LFO options will have the "seed" value too (which wouldn't do anything) but we could add some template logic to hide that.

It's worth noting that you can fake this behaviour by changing the phase of the noise, but that might not always work well.

esnho:
A coordinated noise function can make sense in some usecase, I can try adding a seed value. 
Basically a seed value already exists in the noise function, it's the LinkId where the first letter is used to access a noise buffer, the other letters are used to offset the sampling in order to maximize the output differentiation.

Maybe we can define the seed as enum:
- Auto = -1 -> current behaviour
- 0 to 25 -> select a buffer and disable the offset

I'll check if I'm able to add a dropdown with the seeds during this weekend I guess.

funwithtriangles:
So you think that's better than just a sliding value for the seed? Remember you can also type the value to get it exact (by double clicking). I guess it's nice having auto by default, although with your system, does it not mean that there's a 1/25 chance that two noises will be the same?

src/utils/oneDimensionalNoise/index.js
```diff
@@ -3,31 +3,40 @@
...
+  getValue (x, id) {
     const xFloor = Math.floor(x)
     const t = x - xFloor
     const tRemapSmoothstep = t * t * (3 - 2 * t)
+    const bufferId = id.charCodeAt(0) & BUFFERS_COUNT
+
+    let offset = 0;
+    for (let i = 1; i < id.length; i++) {
+      offset += id.charCodeAt(i)
+    }
```
esnho: There's a 1/25th chance that noises share the same buffer but the offset will differentiate them. :)

esnho:
There's a 1/25th chance that noises share the same buffer but the offset will differentiate them :) 

Check these lines, they calculate an offset using the link id.
https://github.com/nudibranchrecords/hedron/pull/145/files#diff-f32667dbebfa2bd6d3f7d944a0a3c1a1R30

To be honest at the moment I don't know how that noise function could handle a float number as seed...

esnho:
Meanwhile I added the seed dropdown :)

Unfortunately I don't know how to hide it when a non-noise LFO is selected.

dropdown

in action

esnho:
Here a demostration that offsets are hard to spot:

https://editor.p5js.org/esnho/sketches/dHf-zrb9P

funwithtriangles:
Seed dropdown looks great! I'll have a proper look at this when I get the chance and will think about the template logic to hide the dropdown. Thanks again!

funwithtriangles:
This is all fixed up now! In order to "seed" to show only when shape was "noise", I edited the selector that gets the node IDs for each option:

```javascript
/* selectors/getInputLinkLfoOptionIds */

import getNodes from './getNodes'

export default (state, linkId) => {
  const link = state.inputLinks[linkId]

  if (link.input && link.input.id === 'lfo') {
    const optionIds = link.lfoOptionIds
    const optionNodes = getNodes(state, optionIds)
    const shapeOpt = optionNodes.find(node => node.key === 'shape')
    const isNoise = shapeOpt.value === 'noise'

    // Only show "rate" option for shots
    if (link.nodeType === 'shot') {
      const typesForShot = ['rate']
      if (isNoise) {
        // If its noise, also show the "seed" option
        typesForShot.push('seed')
      }
      return optionIds.filter(id => {
        return typesForShot.includes(state.nodes[id].key)
      })
    } else {
      // For params, show all options if noise
      if (isNoise) {
        return optionIds
      } else {
        // If its not noise, omit the "seed" option
        return optionIds.filter(id => {
          return state.nodes[id].key !== 'seed'
        })
      }
    }
  }

  return undefined
}
```

I also had to change the `shouldComponentUpdate` method in the `InputLink` component as that was optimised to only update for specific changes in the app state.

Thanks again for this!

Issue #149, opened 2019-01-30
Ignore node_modules inside sketches folder


funwithtriangles:
Currently Hedron will look inside of all folders in the sketches folder. There's a chance a node_modules folder could be in there, so it would make sense to ignore that.

funwithtriangles:
Fixed https://github.com/nudibranchrecords/hedron/pull/158

PR #157, opened 2019-01-31
finds sketches in subfolders of your sketch folder

cale-bradbury:


cale-bradbury:
Seems to be an issue with fs not playing nicely with the `sinon` or `proxyquire`  in `sketches.spec.js`

cale-bradbury:
Would love if you could check a look at the test file @funwithtriangles still new to unit tests and am wondering if I went about stubbing the fs in a good manner?

funwithtriangles:
Thanks for this! I'll have a look tomorrow. Worth thinking about if there are a ton of folders, such as a rogue node_modules folder in there. Related bug: https://github.com/nudibranchrecords/hedron/issues/149

cale-bradbury:
I think at the moment it doesn't look for subfolders if it finds an index and config, and node_modules would be a child of a folder with a index/config. I added a some node bits to a new sketch I'm working on and nothing seemed to explode

funwithtriangles:
> node_modules would be a child of a folder with a index/config. 

Hmm I'm thinking more if `node_modules` is on the same level as all the other sketch folders. Thinking a simple check for "node_modules" directory name would be a safe way to stop it from diving forever!

cale-bradbury:
I'll add an array of ignored folders with node_modules, perhaps in the future we could let a user add to this via settings

cale-bradbury:
this is merged into #158

Issue #93, opened 2018-09-08
Better Sketch Organization


cale-bradbury:
The add sketch window gets a bit unwieldy as your sketch library increases, I think it would be a good idea to add collapsable grouping. 

The question is if it makes more sense to have an optional `category` param in the sketch `config.js` or to allow your sketch folder to have sub-folders with more sketches, and generate the groups via the sub-folder name.

funwithtriangles:
I think sub-folders make a lot of sense. Organising your sketches is a higher level sort of thing and I feel the mental model of folders is better. The only issue is that the only string for the categories will be the folder names, and if you want to have spaces in the names you'd have to have spaces in your folder name, which feels so wrong.

One other option is in-app organising, where you create folders and the structure gets saved somewhere. One issue I can see with this, however, is whether this structure gets stored globally or per project. I'd imagine this makes sense if you keep your sketch folder the same and it grows. The way I'm working at the moment is a different sketch folder for each show.

cale-bradbury:
Maybe a folder could contain an optional config file with information like name, author, description etc. This would be really good for releasing 'packs' of visuals around a theme

funwithtriangles:
Yeah, I was thinking about an optional config file too. It's a little clunky but probably the best solution and I like the idea of it having extra uses for packs, etc. For simple grouping I'd probably pass on the config file and just use Pascal case strings for folder names (e.g. ParticleSystems, BackgroundEffects, ForegroundEffects).

funwithtriangles:
How would the collapsible grouping preferences be saved? Would it be per project or a global thing? I think the simplest way would be for them to all start closed and then their open/closed state is saved in the project JSON file.

cale-bradbury:
Yeah, that seems like the best way to go

funwithtriangles:
Big improvement with https://github.com/nudibranchrecords/hedron/pull/158

Wondering about folder names as option for categories to save people having to add it to a config? Also I know there are a few other things you want to do @netgrind so leaving this open for now!

funwithtriangles:
Fixed https://github.com/nudibranchrecords/hedron/pull/183

PR #158, opened 2019-02-01
Categorizes Sketches

cale-bradbury:
Adds `Revealers` to the `addSketch` window, organizes sketches based on a `category` tag in it's `config.js`
relates to #93


cale-bradbury:
Feel like this is good for now, Going to work on adding different organization types down the line (author, folder), an after 0.5.0 thing

funwithtriangles:
Thanks for this! I've made a few changes. Quick explanation:

- You didn't need to create new actions/reducer stuff for the opening and closing. There's `UI_AUX_TOGGLE_OPEN` which can be used for any random thing toggled open and closed, as long as you give the action a unique ID.
- To make things simpler, I've made an `AuxRevealer` container, where you just give the container component a unique `auxId` and it takes care of the rest. I've used `AuxRevealer` in the `AddSketch` component
- I've simplified `AddSketch` by moving most of the categorisation logic into a new selector, `getCategorizedModules`, so that the component can stay dumb
- `getCategorizedModules` is quite nice, it returns `looseItems` and `categorizedItems`, so when rendering the component we can have loose items up the top if the author has chosen not to use categories.

Let me know if that all makes sense! If that's all good I'll go ahead and merge. :)

cale-bradbury:
I feel like there was a reason I didn't use UI_AUX_TOGGLE_OPEN, but I don't quite remember, maybe just reducing chance of clashing ids (since I was just using the category name). I'll check this out tonight, but I trust it's good!

funwithtriangles:
Yeah I guess there is a slight namespace concern, but I've prefixed it with `sketchcat_` so I don't think it's too risky :)

cale-bradbury:
I had merged the other pr into this one earlier, haha, oops, just going to close that one and this will merge them both :)

funwithtriangles:
Just saw you were using `indexOf` instead of `includes` which is a bit neater for array checking.

In regards to your tests, they all look fine. I think if we were to be properly rigorous we'd think about how to test the recursive folder check but it's hurting my brain trying to think about how to mock that up with stubs so I'm just going to leave it for now!

Thanks again for this!

PR #182, opened 2019-02-20
made ctrl/cmd + comma open the settings page

cale-bradbury:
a la ableton, ctrl+s is saving so I figure this is a good alternative?


funwithtriangles:
Awesome. Thanks!

PR #183, opened 2019-02-20
AddSketch screen can now organize based on category, author or folder path

cale-bradbury:
folders also display sub revealers if they are nested deep

A big thing that needs to be added in regards to this before .5 is released is for loaded projects to handle new data that may have been added to `setCoreState.js`. Perhaps replacing `uNodeCreate` with a new `uNodeCreateIfMissing` function and just call `setCoreState(state)` again whenever a new project is loaded? Want to get your thoughts @funwithtriangles


funwithtriangles:
Wow, great work Cale! Haven't had time to properly test it but I love the approach of it all.

Yes I am very familiar with this backwards compatibility issue you mention. I think your fix makes sense and is a good way to go. Steps to achieve this:

- Create action (possibly inside of `store/nodes`), could be something like `U_NODES_CREATE_ALL_CORE_ITEMS`, `uNodesCreateAllCoreItems`
- Convert `setCoreState.js` into a listener of this action so that it can easily be triggered. Again, this could be moved into `store/nodes/listener.js`
- Use a `store.dispatch()` to call this action inside of `renderer/index.js` instead of the direct function call that is currently being used
- Create the `uNodeCreateIfMissing` action and associated saga with the logic to check the node before creating it
- Dispatch the `uNodesCreateAllCoreItems` inside `loadProjectRequest` (`store/project/sagas`). I'd dispatch it right after `projectRehydrate` is dispatched (line 60).

Let me know if that makes sense and if you're up for taking it on? Otherwise I'll try and find the time to do it :) Could keep it as part of this PR or as a new one? Let me know what you think. Maybe better to split it out.

One last thing to note: this doesn't completely solve backwards compatibility (e.g. there may be new properties added to param nodes that break things if they don't exist). This needs another approach using if statements where necessary and is something I'm a little reluctant to do before version 1 as I don't want to litter the code base with too much checking logic.

cale-bradbury:
If you are down to take this on that would be great! Have a lot of other things on the plate and it sounds like you have a good idea for going about this.

Maybe in terms of backwards compatibility project files could store their version number, and we could just check that and have a function to upgrade from core version to version, that way only the last release needs to be worried about, could start by saving the number in .5 and worry about implementation for .6

funwithtriangles:
Okay cool, I'll merge this PR once I've checked it and then have a look at backwards compatibility in a separate issue. My only concern is that we might spend a lot of time and effort making sure projects work from version to version, when it might just be quicker to recreate the project from scratch each time a new version comes out. 

Either way I like the version number idea, just not sure if we should slow ourselves down in terms of dev time until the app is a little more mature.

funwithtriangles:
Can continue discussion on backwards compatibility over at https://github.com/nudibranchrecords/hedron/issues/184

funwithtriangles:
Great work Cale! Thanks again.

Note: I did a couple of tiny tweaks on one of the components and added some extra info in the docs. There are a bunch of extra commits because I merged the latest `dev` in to make sure was all still functioning. :)

Issue #198, opened 2019-03-15
Reimporting params breaks app


funwithtriangles:
This is because `sketchId` is not added when reimporting a param. This still causes some `NaN` issue, which needs looking into (it fixes itself if you save and refresh). Might be that the `defaultValue` is being parsed as a string.

Also might be worth thinking of a clean way to reuse the code to create/reimport params rather than the duplicated code we have at the moment.

funwithtriangles:
Fixed by https://github.com/nudibranchrecords/hedron/pull/202

PR #202, opened 2019-03-20
Fix ups for param import from config

funwithtriangles:


funwithtriangles:
Fixes https://github.com/nudibranchrecords/hedron/issues/198

cale-bradbury:
true

PR #206, opened 2019-03-22
Toggle for advance audio features in the settings menu

cale-bradbury:
You can now disable audio texture generation and full audio spectrum computation from the settings menu, may as well get some speed gains if you aren't using them


funwithtriangles:
Hey mate, thanks for this! I've made some changes to the way settings are handled. Instead of doing the logic for settings inside of `inputs/AudioAnalyzer`, I've made a new settings listener. Feel like it's a bit weird to have settings logic next to all that complex audio processing stuff!

When you have time, would it be possible for you to update the docs to explain what these settings do? I'm not entirely sure myself! Would be great to see a working example of how to use these features. :)

**Explanation of code changes:**
In order to achieve this I'm passing the settings into the `update` method of `AudioAnalyzer`. So we have the settings logic all happening in one place, updating the state. Then then the newly updated state gets fed into the `AudioAnalyzer`. It's best to keep to this sort of data flow.

In order to have a listener for settings, I had to create two different types of actions, we've got a `U_SETTINGS_UPDATE` and an `R_SETTINGS_UPDATE`. The `U` is for "user" and `R` is for "reducer". I'm using this pattern elsewhere in the app. Anything that is a `U` action doesn't affect the state directly, but instead is handled by a listener (or a saga, but I'm phasing those out eventually!). An action with an `R` is directly seen by the reducer, so directly updates the state. So the listener handles the user action, and then fires a reducer action after making some changes to the outcome of the payload.

cale-bradbury:
So right now, you can access the audio data as a 1D texture in I think that you can get it by importing `AudioAnalyzer` into your sketch and accessing it's `.texture` (for passing through to shaders) or `.data` (for accessing each individual sampled frequency), however, it appears that on my own branch I am adding it to the `allParams` that gets passed through the sketch `update` function. 

I could add that to this branch, but maybe you have thoughts on a better way to make these kind of global objects accessible to a sketch @funwithtriangles

funwithtriangles:
@cale-bradbury We could either add this as an argument to the relevant methods or perhaps just expose it as a global variable?

Currently we have `window.HEDRON` which just has a `dependencies` property. But we could maybe we could also have a `data` property and put this in there? From my understanding we just need a single reference to it so it doesn't actually needed to be handed over on every update.

So something like this:

```javascript
const data = window.HEDRON.data

class MySketch {
   ...
   update () {
      this.thingUses(data.audioTexture)
      this.otherThingUses(data.audioData)
   }
}
```

What do you think?

src/engine/index.js
```diff
@@ -257,6 +257,7 @@ export const run = (injectedStore, stats) => {
         sketch = sketches[sketchId]
         const params = getSketchParams(state, sketchId)
         allParams = getSketchParams(state, null, sceneId)
+        allParams.audioTexture = store.getState().inputs.texture.value;
```
cale-bradbury, line 260: @funwithtriangles Don't know where/if we landed on this, do we want to just toss the texture in `allParams` for now? Could potentially create an `assets` object too with a bit of a mind towards an asset library that we could just let sketches have full access to . Love the idea of sketches being able to just create weird datamash stuff by swiping from all the textures, models, etc.

src/selectors/getNodeInputOptions.js
```diff
@@ -16,6 +16,12 @@ export default (state, nodeId) => {
       label: 'Audio',
       exclude: ['linkableAction'],
     },
+    {
+      value: 'texture',
+      type: 'texture',
+      label: 'Texture',
+      exclude: ['linkableAction'],
```
cale-bradbury, line 23: Maybe we want textures to be linkable? A little out of my depth with this system, was mostly just copying audio to get the audio texture into the store

src/inputs/AudioInput.js
```diff
@@ -8,13 +8,16 @@ export default (store) => {
...
       input.normalizeLevels = state.nodes['audioNormalizeLevels'].value
       input.levelsFalloff = Math.pow(state.nodes['audioLevelsFalloff'].value, 2)
       input.maxLevelFalloffMultiplier = 1 - Math.pow(state.nodes['audioNormalizeRangeFalloff'].value, 3) * 0.01
       input.smoothing = state.nodes['audioLevelsSmoothing'].value * 0.99
       input.levelsPower = state.nodes['audioLevelsPower'].value * 3 + 0.5
-      bands = input.update()
+
+      bands = input.update(state.settings)
       store.dispatch(inputFired('audio', bands, { type: 'audio' }))
+      store.dispatch(inputFired('texture', input.texture, { type: 'texture' }))
```
cale-bradbury, line 20: need to get this to be assigned to `input.texture.audioTexture.value` with texture being a dictionary rather than the texture object. correct solution being to make a `texturedFired` function?

src/store/inputs/reducer.js
```diff
@@ -7,6 +7,10 @@ const defaultState = {
     value: 0,
     assignedLinkIds: [],
   },
+  texture: {
+    value: null,
+    assignedLinkIds: [],
+  },
```
cale-bradbury, line 13: One thing of note here, opening up old project files, if they are missing any attributes in the default state, they just continue on without them, need to do some sort of `state = {defaultState, ...loadedState}`

src/engine/index.js
funwithtriangles: Thinking it might be better just to expose this as a global? As mentioned above? Seeing as this isn't data that relates to a sketch and is more of a global function? Unless we plan on having this modified per-sketch.

```javascript
const data = window.HEDRON.data

class MySketch {
   ...
   update () {
      this.thingUses(data.audioTexture)
      this.otherThingUses(data.audioData)
   }
}
```

src/selectors/getNodeInputOptions.js
funwithtriangles: A a `linkableAction` is any button that can be connected with inputs. So for instance, the "enable" button for an LFO is linkable. The only inputs that these are enabled for are user controlled MIDI. So you were right to exclude it for audio texture, it definitely wouldn't make much sense!

funwithtriangles: Although I'm a bit confused as to how this can be assigned to a param, is it a 1D texture?

src/store/inputs/reducer.js
funwithtriangles: 100% this

Issue #184, opened 2019-02-20
Backwards compatibility


funwithtriangles:
As the app changes, project files can suddenly break everything if certain items are missing specific properties. At some point we'll want backwards compatibility so that this doesn't happen from version to version. One easy fix is for missing top level "core nodes" (e.g. A/B mix value) to be added if missing from a project. This has been discussed over at PR https://github.com/nudibranchrecords/hedron/pull/183. 

There are other issues involving deeper changes to node structures that would need a little more traversing to fix. One solution would be to save the version number of the app into each project and check against it on load. If the version number doesn't match, an upgrade procedure is carried out to add any missing parts to certain nodes. This idea is nice but could be a lot of work to maintain, with quite a bit of room for error.

I think at this stage in the process it might make sense to keep backwards compatibility to a minimum and accept that old projects are going to break, until we're closer to version 1 and have some more of the core features in place. Hopefully by then the breaking changes from version to version might be a little less drastic.

funwithtriangles:
After some thought, I can see the advantages of having project upgrade functionality at these early stages of the app.

The strategy of updating the core nodes (https://github.com/nudibranchrecords/hedron/pull/183#issuecomment-465558042) may not be necessary if we're writing version update functions for each version, as this piece of the state can be added here instead. What's nice about this is that it's very easy to see how the state structure changes from version to version. Here's my proposed method:

- Projects store the current Hedron version when saving
- If an old project is loaded into a newer version of Hedron, a warning appears for the user. A backup of their project is automatically saved (e.g. `myproject.backup-v0.4.1.json`) and if the user proceeds, the project is updated.
- The update procedure will be a series of functions. Every time a new version of Hedron is released, there needs to be an update function that comes with it. Simple example of what this might look like:

```javascript
// upgraders/0.5.0.js
export default project => {
  // Add sketch organisation core node if it doesnt exist
  if (project.nodes.sketchOrganization === undefined) {
    project.nodes.sketchOrganization = {
      title: 'Sketch Organization',
      type: 'select',
      value: 'category',
      id: 'sketchOrganization',
      options:[
        {
          value: 'folder',
          label: 'Folder',
        },
        {
          value: 'category',
          label: 'Category',
        },
        {
          value: 'author',
          label: 'Author',
        },
      ],
    }
  }

  return project
}
```
[...]

PR #215, opened 2019-03-29
Feature/audio increment

cale-bradbury:
Adds an increment option to audio modifiers, this is really good for rotation, or any other parameters that loop. causes the parameter value to continuously shift forward/backwards with the specified frequency band.

One issue is saved projects will be missing this new parameter, wondering what your thoughts are @funwithtriangles I remember at one point talking about possibly creating an upgrade function to funnel projects through, version to version


funwithtriangles:
This is fantastic Cale! Have thought about this kind of feature before but never got round to implementing. Why not have it work for LFO too?

Just added my curent thoughts on backwards compatibility: https://github.com/nudibranchrecords/hedron/issues/184#issuecomment-477952552

cale-bradbury:
Yeah, could add it to LFO as well, one thing that's a little tricky with this is that you often need super small numbers for the upper range, I suppose this is okay as you can type a value in though?

funwithtriangles:
Hmm I guess that just depends on your sketch also? I don't see the harm in having it for LFO as well as audio, might be useful for someone!

cale-bradbury:
I could imagine something pulsing forward in a sine motion feeling v nice ~~~

funwithtriangles:
Moving any new pull requests to 0.6.0 apart from bugs, so 0.5.0 can get released

funwithtriangles:
I converted this over to a modifier (had to rework modifiers a bit for this to work). The reasoning behind doing this is that eventually I'd like modifiers to be dropped in like sketches, so I want them to be flexible and allow for stuff like this increment feature. The API is a bit clunky for adding modifiers at the moment but I still think its neater to have them like this. :)

It also means that making this work for both audio and LFO is as simple as adding "LFO" to an array.

Issue #11, opened 2018-02-22
Post processing of shaders


funwithtriangles:
It would be good to have final shader effects after the scene has been rendered. Using something like the [EffectComposer](https://github.com/mrdoob/three.js/blob/be6bd29fd1461c5c6158abefed385271a386117b/examples/js/postprocessing/EffectComposer.js). 

Should be able to:

- GLSL files loadable like sketches
- Treat each shader like a sketch, where uniforms can be controlled like params
- Add/remove/reorder shaders

Bugs to avoid: 

- Current hacked version goes very wrong if you load in a new project without restarting the app. Guessing something to do with shaders still being loaded in the main rendering pipeline and competing with each other when they're loaded again.

To dos:

- [x] Create nice example sketch to test (space scene with objects, lights, etc)
- [x] Fix resolution issues
- [x] Bloom sketch in the example
- [x] Color grading sketch in the example
- [x] Simple Feedback sketch in the example
- [x] Hot reloading (ensure it's not loaded multiple times)
- [x] Hot reloading (ensure changes happen to post stuff)
- [x] Adding
- [x] Removing
- [x] Sketch reordering
- [x] Scene reordering
- [x] Only reset composer once when a scene is deleted with lots of sketches
- [x] Remove associated render targets when sketch is deleted
- [x] Test if updates work when scene isn't active
- [x] Add visual indication for global post scenes
- [x] Fix project settings issue
- [x] Ensure composer is reset on project load
- [x] Example project.json
- [x] Sketches with postprocessing related methods should display differently (in the add sketch screen and in the usual places)
- [x] Dev docs
- [x] User docs
- [x] Docs: Globals section
- [x] Disable sketch post processing (save for later: https://github.com/nudibranchrecords/hedron/issues/310)

Bugs: 

- [x] Feedback effect resets when switching scenes or resizing preview

funwithtriangles:
@netgrind Do you think post processing should be per scene, or at the very end of the rendering pipeline, so it works across scenes? I'm leaning to per scene, which fits better with your GUI idea of having the shaders underneath sketches on the right.

cale-bradbury:
There are really good pros and cons to both, I'd say scenes as that is the most versatile for the artist at the potential pre-show cost of re-setting up shaders across scenes and performance cost of doubling shaders while transitioning. Could also add a second global pass in a later milestone

funwithtriangles:
[...]

PR #311, opened 2019-10-16
Feature/postprocessing

### LINKED ISSUE #260: Sketches without params or shots defined shouldn't cause issues
Reimporting a sketch with no shots or params defined in the config causes issues. Perhaps the easiest way to fix this is to make these default to an empty array.

### LINKED ISSUE #297: Consider how best to do global vars
Global vars seem to be the most convenient way to expose useful things to sketch authors. At the moment, we have `THREE`, `POSTPROCESSING` and `GLSLIFY`.

I'm also considering adding `TWEEN` for tween.js functionality. Eventually we might also want `HEDRON` for hedron specific utils (such as a `Sketch` class that can be extended, if we need to do this one day).

I'm slightly concerned we're really polluting the global namespace with all of the above and it gets hard to know what is available to a developer.

One simple approach would be to only expose one global var, `HEDRON`, and attach everything to that. 

```javascript
const { THREE, POSTPROCESSING, TWEEN, GLSLIFY, Sketch  } = HEDRON
```

**Pros:**
- Less namespace pollution, just one global var to avoid
- Easily console loggable by devs to see what's actually in there

**Cons:**
- One more line of code to write in order to access the variables
- Might be harder to set VSCode intellisense to recognise global var if its in an object?

funwithtriangles:
First step for post processing. Only global post processing working at this point, but the API is in place. 

@cale-bradbury I'd love to hear some feedback. The best starting point is to load up the `post` example project and play around. After that it would be great if you can take some pre-existing sketches you may have worked on and see if they can be converted without any need for hacks.

The docs for development are updated, but not for the user guide yet.

Previous discusion here: https://github.com/nudibranchrecords/hedron/issues/11
Incidentally fixes https://github.com/nudibranchrecords/hedron/issues/297, fixes https://github.com/nudibranchrecords/hedron/issues/260

To Dos:
- [ ] User guide
- [x] Use object for arguments (e.g. `initiatePostprocessing ({ composer })`)


docs/dev/index.md
```diff
@@ -239,6 +239,60 @@ class Solid {
...
+    // Add the bloom effect as a new pass to the composer
+    const pass = new EffectPass(null, this.bloomEffect)
+    composer.addPass(pass)
+
+    // Return the pass that needs to be rendered to the screen
+    return pass
+  }
+
+  // This method will be called every frame, just like the usual update method
+  updatePostProcessing (p) {
```
cale-bradbury, line 308: Curious about why this doesn't share the standard update?

docs/dev/index.md
```diff
@@ -239,6 +239,60 @@ class Solid {
 module.exports = Solid
 ```
 
+## Post Processing
+Custom post processing, such as pixel shaders, are handled inside of sketches. Hedron's post processing system is a thin wrapper around the [postprocessing](https://github.com/vanruesc/postprocessing) library, so it's a good idea to understand how that works. Multiple passes can be added to the rendering composer using `initiatePostProcessing` and things are updated every frame using `updatePostProcessing`.
+
+### Simple example
+Below is a simple example of how to achieve a bloom effect. A config file is also needed, which is exactly the same as a normal sketch config file.
```
cale-bradbury, line 280: I think it might be a better call to just show how to create the most bare-bones glsl invert, I feel like this would be the most streamlined way of getting someone to their creative sandbox. 

This feeling is enhanced by my thinking that a bloom effect would be a perfect candidate to just have as a sketch we just provide out the gate.

docs/dev/index.md
```diff
@@ -239,6 +239,60 @@ class Solid {
...
+
+```javascript
+// POSTPROCESSING is a global variable available to Hedron sketches
+const { EffectPass, BloomEffect, BlendFunction, KernelSize } = POSTPROCESSING
+
+class Bloom {
+  // Here we add our passes to the composer
+  initiatePostProcessing (composer) {
```
cale-bradbury, line 288: What are your thoughts on having the composer manage the effects rather than the sketch manage the composer? I feel this would help a lot when it comes to layer re-organization and deleting/adding sketches. 

This does mean that doing multipass stuff would be a bit harder, but I feel that it could take creative experimentation to the next level by allowing sketches to fire up "child-passes" that can be reordered in the stack at will.

I've personally had a lot of success coming up new visualizer styles by trying random effects in between a feedback and it's save pass.

docs/dev/index.md
```diff
@@ -239,6 +239,60 @@ class Solid {
...
+module.exports = Bloom
+```
+
+### Other examples
+There are plenty of other examples that can be found in the [example sketches folder](../../example-projects).
+
+### How to see the output in Hedron
+Currently, post processing only works on the final output of Hedron (there are plans for a per-scene option too). To see the output of your passes, the sketch must be added to a scene. This scene must have "Global Post Processing" enabled under the scene settings for the passes to take effect. An icon will appear on the scene thumbnail if this setting is enabled. The scene does not need to be added to any channel, `updatePostProcessing` will always be running with this setting on.
+
+As a convention, it makes sense to have a post processing scene, with post processing related sketches added to it. This scene does not need to have any 3D objects in it and never needs to be added to a channel.
```
cale-bradbury, line 325: It's interesting that you can make multiple post processing scenes, or have it be a mix of 3d and post. Something about it feels weird, but I really like that this lets the user do whatever they want, and could lead to adding effect stack mixing pretty easily

example-projects/post/sketches/space/index.js
```diff
@@ -0,0 +1,94 @@
...
+      .setPath(`${__dirname}/assets/`)
+      .load([
+        'px.jpg', 'nx.jpg',
+        'py.jpg', 'ny.jpg',
+        'pz.jpg', 'nz.jpg',
+      ])
+
+    // Hacky way to get orbit controls working with the viewer
+    const viewerEl = document.getElementById('viewer')
+    this.controls = new THREE.OrbitControls(world.camera, viewerEl)
```
cale-bradbury, line 20: beautiful, maybe we could expose the element in some way for _**advanced users**_, that would save some (future)old sketches if the viewer element ever changes in some way

src/engine/renderer.js
```diff
@@ -30,6 +33,40 @@ export const setRenderer = () => {
...
+  mainPass.renderToScreen = true
+  composer.addPass(mainPass)
+
+  const passes = []
+
+  stateScenes.forEach(scene => {
+    scene.sketchIds.forEach(sketchId => {
+      const module = engine.sketches[sketchId]
+      // Do any postprocessing related setup for this sketch
+      if (scene.settings.globalPostProcessingEnabled && module.initiatePostProcessing) {
```
cale-bradbury, line 57: curious to know what your thoughts are on having effect sketches extend a base class with a flag so we can detect it is an effect (type checking is typescript only right? little rusty) and possibly some utility functions/params?

src/renderer/globalVars.js
```diff
@@ -0,0 +1,26 @@
...
+window.HEDRON = {
+  // Third party dependencies exposed globally for sketch development
+  dependencies: {
+    // Declaring THREE as a global var, so that sketches can use the same instance of three.js as Hedron does
+    // This keeps the library versions matched and also prevents strange things from happening when the library
+    // code is being read from different sources
+    THREE: {
+      ...THREE,
+      // For convenience, also requiring some common three extras
+      GLTFLoader,
```
cale-bradbury: Not needed for this pr, but it might be handy to have as many loaders as we can get our hands on, will help lead way to an asset importing/library type thing

src/store/scenes/listener.js
```diff
@@ -59,7 +61,9 @@ const handleSceneDelete = (action, store) => {
 
   // Delete sketches
   scene.sketchIds.forEach(sketchId => {
-    store.dispatch(uSketchDelete(sketchId, p.id))
+    // We're skipping postprocessing reset to stop it happening multiple times per sketch
+    // Instead it is called once after scene is deleted
+    store.dispatch(uSketchDelete(sketchId, p.id, { skipPostProcessingReset: true }))
```
cale-bradbury, line 66: maybe changing this so the reset function schedules a clean, and if a clean is already scheduled it doesn't do anything

cale-bradbury (commented):
This is looking really good, took the tests for a spin and it felt great, have yet to load up my own but I have no worries. Brought a tear to my eye when I saw the sketch re-ordering. Lots of little thoughts, thanks for asking for my feedback :)
Really great work here, I'm super grateful and joyous that a nice and clean shader system is going to get out to folks, this is really going to take hedron to a new place!

docs/dev/index.md
funwithtriangles: I originally had it inside the update method and can't remember if I came across a specific issue and changed it, or just changed my mind. The main thing I can think of is over the order of updates. The post processing should happen after all 3D scene updates, and having this stuff happen in a separate method ensures that. It also means the user could do something special inside this method that they know will be called last. Perhaps the naming could be improved?

docs/dev/index.md
funwithtriangles: Good thinking! Although the reason I chose the bloom is because its simple to implement via the `postprocessing` library. Either way, having some bare bones example like invert definitely makes sense.

docs/dev/index.md
funwithtriangles: I'm a little confused by what you mean here. Do you mean each sketch should be a single pass and it's up to the user in Hedron to add them and order them, etc? Would be interested to hear more about this.

docs/dev/index.md
funwithtriangles: One reason for allowing this is that you might have some 3D object in the scene that directly relates to the postprocessing effect. Example of this here, where a mesh is being used as a mask for the pixelation:

https://vanruesc.github.io/postprocessing/public/demo/#pixelation

example-projects/post/sketches/space/index.js
funwithtriangles: Heh, I've been doing this for a while now! Ultimately this isn't ideal because eventually all of the rendering may get moved to a different thread where there is no access to the DOM and this tricky will no longer work. Not anytime soon though. I won't ever be changing the ID of that element.

src/engine/renderer.js
funwithtriangles: Hmm what is the advantage of doing that? Having the "global post processing" setting is more of a future proofing thing. It needs to be checked at the moment because we only have global post processing (where the effect works on top of any crossfader stuff). Later on we'll have a scene specific post processing (so will only affect the channel as opposed to the whole fade mix) .

src/renderer/globalVars.js
funwithtriangles: Definitely

src/store/scenes/listener.js
funwithtriangles: I see what you mean, although your approach may take a bit of refactoring. Happy for you to go ahead and make this change if you feel it's tidier. :)

funwithtriangles:
I'm merging this for now in the knowledge that the API will probably change and improve as this feature evolves.

PR #337, opened 2019-11-23
String valueType

cale-bradbury:
Heya, went ahead and tried to add a string type, ran into a few issues which I'm hoping you might be able to shed some light on. I'll comment through the PR with the problem areas


example-projects/text/textBasic/index.js
```diff
@@ -0,0 +1,78 @@
+const THREE = require('three')
+const fontJson = require('../../fonts/droid/droid_sans_regular.typeface.json')
+
+class TextBasic {
```
cale-bradbury, line 4: this sketch and config will need some work before getting added, but getting this up for you now so you can make a string do something right away

src/valueTypes/StringValueType/index.js
```diff
@@ -0,0 +1,15 @@
...
+export class StringValueType extends BaseValueType {
+  defaultValue = 'text'
+
+  Component = ParamStringContainer
+
+  doesValueMatch(value) {
+    return typeof value === 'string'
+  }
+
+  compatibleInputs = {
```
cale-bradbury: I get the following error, but only the first time hedron is compiled, if I cmd+r to reload it begins to work. Though sometimes after that, upon reloading again it freezes
```
TypeError: Cannot read property 'inputs' of undefined
    at processDevices(MidiInput.js ? 6582 : 68)
    at runCallEffect(/Users/cale.bradbury…nternal / proc.js: 524)
```

I'm wondering if this is due to the fact that this type has no inputs?

src/valueTypes/StringValueType/component.js
```diff
@@ -0,0 +1,42 @@
...
+  const store = useStore()
+  const [value, setValue] = useState()
+
+  useUiEventListener('slow-tick', () => {
+    const newValue = store.getState().nodes[nodeId].value
+    if (value !== newValue) setValue(newValue)
+  })
+  return (
+    <TextField
+      value={value || ''}   //This seems to not handle multiple fast keypresses, only updating with one of the keys
```
cale-bradbury: Is this the correct way to be feeding the value into the component? as the comment states, it works well enough until you start typing fast

src/valueTypes/StringValueType/index.js
cale-bradbury: Also forgot to mention that for the second freeze I get the following in the terminal
```
[19645:1123/151459.614385:ERROR:gles2_cmd_decoder.cc(18470)] [.DisplayCompositor]GL ERROR :GL_INVALID_OPERATION : glCreateAndConsumeTextureCHROMIUM: invalid mailbox name
```
Hedron does render, but is very quickly frozen

src/valueTypes/StringValueType/component.js
cale-bradbury: I just had the thought that perhaps why it wasn't handling fast typing is because it was updating the geometry with each change, which could have cause slight lag/frame-drops (especially as your text gets longer)

example-projects/text/textBasic/index.js
funwithtriangles: Yeah makes sense. Perhaps we could provide a few fonts in the global `HEDRON` object. Otherwise I'd advise to have the font located inside the sketch folder itself. Also, worth noting your folder structure is a bit wrong for this example project, see the other ones to see what I mean. Although I was recently thinking we could change the structure of all this anyway: https://github.com/nudibranchrecords/hedron/issues/341

funwithtriangles: It's probably worth doing an upstream merge to this branch so that you're up to date. You shouldn't be pulling in `THREE` like that with the new hedron, it's now a global object. `window.HEDRON.dependencies.THREE`

src/valueTypes/StringValueType/index.js
funwithtriangles: Strangely, I'm not getting these errors. Might be a node issue? You shouldn't need the `compatibleInputs` as it's empty in the base class for value type.

funwithtriangles: I think you might be running multiple instances of Hedron and getting these issues. Check your task manager. Sometimes happens while developing!

src/valueTypes/StringValueType/component.js
funwithtriangles: I think we might want to have some sort of debounce option for the text field. We kind of have this with the `slow-tick` event listener though. My advice would be to get it working with canvas on a plane first, before going for full 3D text.

funwithtriangles:
Hey, I just did a tiny bit of linting and a few little minor edits just as I was looking through it locally. 

So far so good though! Most of my feedback is inline

src/valueTypes/StringValueType/index.js
cale-bradbury: I'll check in tonight and let you know, thanks for the thing to look into!

src/valueTypes/StringValueType/component.js
cale-bradbury: I actually believe the 3d lag can be mitigated by having a sketch that generates all the character geometry on creation then re-assembles into a parent object container as the text updates. Having a debounce, or not applying the text until enter is pressed could be good though

src/valueTypes/StringValueType/index.js
cale-bradbury: No more errors :) I didn't see those again, I'll keep an eye open for multiple processes in the future though!

funwithtriangles:
Hey @cale-bradbury, how's this one going? Have you finished on your end and want me to take a look?

funwithtriangles:
Hey @cale-bradbury I went ahead and fixed up some things. It's ready to go now. I'm merging it in, thanks so much! Explanation of changes below.

**Lagging text input**
The reason for the lag on the text input is that input changes are dispatched to update the state, the component is notified of this and then uses the new state to update itself. This is a "unidirectional data flow" which keeps things nice and neat when you know the update is only coming from one place, especially as params can be manipulated from multiple places (e.g. inputs, macros and shots). 

In a normal React/Redux application, this update of the state would happen instantly and notify the component, but for performance reasons I'm bypassing the whole thing, checking the state and updating the UI on a "slow tick". Param bars don't actually update at 60fps, but something like 30fps instead. This works fine for these sorts of inputs, as the GUI doesn't need to be frame perfect for the user.

However with text it's different, If you're typing super quick, you can beat this tick before the data has updated, and when it does you're suddenly back a few keystrokes. So I'm cheating a bit and updating the text input from _within_ the component, rather than doing the unidirectional approach. We're still _also_ updating on the tick, in case the update is coming from a macro or a shot.

**Example sketch**
I cleaned this up a little bit. I've gone for `MeshBasicMaterial` so lighting is not to see colour changes (this was confusing me a lot because of the emissive colour!). I've also removed the bevel options as these were only visible when the text change and also causing performance issues. As this is just an example sketch I think it makes sense for it to be as intuitive and simple as possible.

PR #349, opened 2020-01-16
Feature/postprocessing per scene

### LINKED ISSUE #281: Fragment shader example sketch
Probably possible using new post processing feature. Would be good to get uniforms automatically generated from the params using a loop.

### LINKED ISSUE #341: Change structure of example projects
Currently looks like this:
```
projectA/
   project.json
   sketches/
      fooSketch
      barSketch
projectB/
   project.json
   sketches/
      wooSketch
      waaSketch
```

Instead we could just have:
```
projectA.json
projectB.json
sketches/
   fooSketch
   barSketch
   wooSketch
   waaSketch
```

### LINKED ISSUE #351: Easier access to the output resolution
Currently you need to get the output resolution from the renderer. Would be nice if this was a property accessible directly from `constructor` and `update`!

funwithtriangles:
Post processing is no longer limited to global effects, by default it will now only affect the scene it is happening inside of! Of course, you can still check the "global post processing" option for it to affect all scenes.

To Do:
- [x] Try using enable/disable on all scene passes rather than adding removing on the fly
- [x] Dont do a save pass unless there have been other passes, otherwise just use a render target on the render pass. 
- [x] May not even need to do a render pass if there's no `root` object in any sketch of that scene? Thinking if a sketch simply has a pass for a full screen shader sort of thingy
- [x] Test global effects are still working properly
- [x] Test all updates (add/delete scene, add/delete sketch, reorder scene, reorder sketch, set global post)

Fixes #351 
Fixes https://github.com/nudibranchrecords/hedron/issues/281
Fixes https://github.com/nudibranchrecords/hedron/issues/341


funwithtriangles:
I'm seeing how things go with a single `update` method instead of the extra `updatePostprocessing` method. There might be some unforeseen downsides to having all updates inside of a single method, but it definitely feels a lot less clunky!

funwithtriangles:
@cale-bradbury @amsXYZ would love your feedback on this :)

One big difference is there is no longer access to `composer`, you just return an array of passes you want added in. Wondering if there is any use cases where the user might still want access to the composer?

funwithtriangles:
Thinking that `initiatePostProcessing` should have similar arguments to the constructor

cale-bradbury:
I love the passing of effects as an array instead of dealing with the composer yourself. I don't see any immediate need, it could always be added to the `window.HEDRON.data` object if someone finds a need for it

src/engine/renderer.js
```diff
@@ -260,11 +315,11 @@ export const render = (sceneA, sceneB, mixRatio, viewerMode, deltaIn) => {
       outputCanvas.style.display = 'block'
 
       // Render for output
-      renderLogic(sceneA, sceneB, mixState, mixRatio)
+      renderLogic(mixState, mixRatio)
       // Copy pixels to output canvas
       copyPixels(outputContext)
       // Render for preview
-      renderLogic(sceneA, sceneB, viewerMode, mixRatio)
+      renderLogic(viewerMode, mixRatio)
```
cale-bradbury, line 341: I wouldn't let this block this branch, as it's the same amount of computation as before. But I'm wondering your thoughts on scaling down the output texture and rendering that to the preview @funwithtriangles ? I feel that would save processing when you have both the output and preview running during a show, preview might look a little chunky but I don't think that's really a big deal

funwithtriangles: I think copying the pixels is actually faster, no?

cale-bradbury: yeah, but renderLogic gets called twice in this branch of the if statement, once for the preview and once for the output, which re-renders the whole scene yeah?

funwithtriangles: Heh, this is actually a bit confusing. This block is for when you have the viewer set to something different than the output. For example, you may be previewing channel A but the audience is seeing the mix, which is set to channel B, therefore two renders! Not actually a feature I ever use but I think it's a nice thing!

funwithtriangles: The nice thing is, that if you have viewer mode set to 'A' and the mix is all the way over to 'A', it will just render once.

funwithtriangles: However I guess in this instance it would make sense to render the preview at a much smaller resolution. Which is probably what you were trying to tell me! Hah sorry long day 😅

funwithtriangles: I've made an issue for it https://github.com/nudibranchrecords/hedron/issues/352

cale-bradbury: Oh, I thought the two calls were rendering the same scene, looks like you gathered a different improvement though, haha.

I didn't know you could set the viewer to something else, I'll have to play around with that!

funwithtriangles:
Fixed an issue where the passes weren't actually updating when channels were being swapped around. This seems to work fine now but I'm slightly concerned about performance when adding a scene to a channel.

Not sure if adding/removing composer passes on the fly is the issue or not. Perhaps it makes more sense to add passes to the composer for _every_ scene and then just enable/disable as needed. Rather than adding/remove.

funwithtriangles:
So I've reworked this so that all passes from every scene available are added to the composer initially. When a scene is added to channel A or B, it simply enables/disables the relevant passes and sets the output texture for that channel.

This method has less tricky edge cases and is probably more performant when swapping scenes as no initialisations are happening.

PR #372, opened 2020-02-19
Custom sketch render methods.

amsXYZ:
This PR explores custom render methods per sketch, which allow sketches to have full control of the render loop (through the custom ``CallbackPass``, instead of running a simple ``RenderPass``).

Work needs to be done in the following areas:
- Pass depth and stencil buffers to these passes.
- Make sure the screen buffers are not cleared between ``CallbackPasses`` (unless you're explicitly doing it so inside the render callback).
- Render order: Should these be rendered before or after the main ``RenderPass``? 
  - How should the render order be defined if you have sth like this?
    - SCENE
      - RootSketch
      - RenderSketch
      - RootSketch
      - RenderSketch

Hopefully we can figure out these details together ;)

### DIFF
--- a/src/engine/Scene.js
+++ b/src/engine/Scene.js
@@ -5,6 +5,7 @@ class Scene {
     this.scene = new THREE.Scene()
     this.camera = new THREE.PerspectiveCamera(75, null, 1, 100000)
     this.camera.position.z = 5
+    this.renderMethods = new Map()
   }
 
   setRatio (ratio) {
--- a/src/engine/index.js
+++ b/src/engine/index.js
@@ -152,6 +152,7 @@ export const addSketchToScene = (sceneId, sketchId, moduleId, shouldSetPost = tr
 
   sketches[sketchId] = module
   module.root && scene.scene.add(module.root)
+  module.render && scene.renderMethods.set(sketchId, module.render.bind(module))
 
   if (shouldSetPost) renderer.setPostProcessing()
 }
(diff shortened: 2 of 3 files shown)


funwithtriangles:
Great stuff, thanks! In terms of render order, I'm inclined to think it should happen after the main render pass. Do note that if there is no `root` defined, there won't even be a render pass. This is useful if someone is just doing post processing and has a bunch of passes on a sketch with nothing else. I'm assuming these sorts of passes will always happen after this render method?

funwithtriangles:
Great stuff, thanks! In terms of render order, I'm inclined to think it should happen after the main render pass. Do note that if there is no `root` defined, there won't even be a render pass. This is useful if someone is just doing post processing and has a bunch of passes on a sketch with nothing else. I'm assuming these sorts of passes will always happen after this render method?

amsXYZ:
Yeah, but what I meant is that all "regular" sketches (the ones with a root object) will be rendered in one single render pass (either **channelA** or **channelB** scene), so the order of sketches in a scene wouldn't be preserved and this could be confusing...

Only solution to this I see is to give each "regular" sketch is own scene (at the end of the day, this is just used as a container for geometries, lights, etc.) and render all of them with the channel camera. Only issue that I have with this is that, at this point, we might drop root support as well... But I'm open to keep discussing.

funwithtriangles:
I see what you mean but I think having both root style sketches and render style sketches in the same Hedron scene is a bit of an edge case, so we can afford the slight confusion over the fact that sketch order won't be completely respected. As long as it's well documented I don't think it's too much of an issue.

I don't think having a separate scene for each sketch is that workable. For instance, I like to have a sketch called "env" which controls background colour, fog and basic lighting. I haven't tested it but I'd imagine at least the fog here wouldn't behave properly.

PR #396, opened 2020-09-10
Windows 10 Dark Mode Fix

cale-bradbury:
Unlinks React Dev Tools so that Electron doesn't freeze when booted up while Windows is in dark mode
see https://github.com/electron/electron/issues/19468


funwithtriangles:
It looks like the best fix for this might just be updating to the newest version of Electron

funwithtriangles:
Closing in favour of #402. I've tested in Windows dark mode and all seems well! Seems like a nicer solution to just upgrade packages :)

PR #448, opened 2024-09-29
[WIP] Save load json

### LINKED ISSUE #449: Save/load JSON
- [x] json file load input
- [x] main thread start sketches server
- [x] render thread replace project data
- [x] refactor other main/renderer communication using `invoke`
- [x] abstract renderer code into a util (e.g. `startSketchesServer(dir).then(....)`)
- [x] do same for opening sketches folder
- [ ] very basic error alert system (just wrap `window.alert` for now)
- [x] save using system menu
- [x] save as using menu
- [x] load using menu
- [x] store sketches dir string in project data
- [x] load sketches dir from project file (or use neighbour)
- [ ] allow users to manually select sketches folder if project isn't linked
- [x] main thread tidy code into handler files
- [x] FIX: when loading the sketches dir for the first time, it's always going to be returned as absolute, but we'd rather save it as relative when possible. It can be saved in the store in whatever format, we can handle this with some logic when saving the project. Upon the next load, it will stay as relative with no issues.

- [x] Also: sketches folder as a sibling/sibling-child
When a user saves a project next to their sketches folder, it should be stored in the project as a relative URL rather than absolute. This should also happen if the sketches folder is somewhere lower down than the project file (e.g. neighbouring directory to project.json). This means that when we load we don't actually have to go looking for `sketches` as a neighbour. We just know it's a relative URL and find it with the path given.

funwithtriangles:
Saving and loading done, just a few little details i wanna add/clean up before merging


src/main/index.ts
```diff
@@ -76,13 +79,69 @@ app.on('window-all-closed', () => {
...
+      try {
+        // Read and parse the JSON from the project file
+        const fileContent = fs.readFileSync(projectFile, { encoding: 'utf8' })
+        const projectData = JSON.parse(fileContent)
+
+        // Get the directory of the project file
+        const projectDir = path.dirname(projectFile)
+
+        // Path to the neighboring 'sketches' directory
+        const sketchesDir = path.join(projectDir, 'sketches')
```
cale-bradbury: Maybe the sketches path(s) are stored in the projectData?

example-project/project.json
```diff
@@ -0,0 +1,168 @@
+{
+  "sketches": {
+    "c64ceed4673": {
+      "id": "c64ceed4673",
+      "moduleId": "solid",
+      "title": "Solid",
+      "paramIds": [
+        "64ceed46737",
```
cale-bradbury, line 8: A few thoughts on the file format

1. Do we need id when we can just use the objects key?
2. Will paramIds ever be shared? Kind of feel like the file would be more readable if this was just a `"params"` array, and the param objects (starting on line 37) also just contained their value

I'm assuming that this current layout is close to how it is stored in memory? Think it could be worth having a de/serialize step so the file could be more readable in commits, or if they ever need to be touched by hand

cale-bradbury (commented):
few thoughts, looking good though, will be nice to have this back up and running

src/main/index.ts
funwithtriangles: So my plan is that the path will definitely be stored in the project data. However I think there should be two possibilities when the user saves the file:
- if upon saving, the sketches dir is in the same directory as the project json even a child directory: save as a relative path
- Otherwise, save it as an absolute path

When loading, we check to see what kind of path it is and locate the folder accordingly. This means that if you're keeping your save files next to your sketches folder, they can be moved around and shared without having to relocate the sketches directory when opening the project in hedron.

But if you save your project somewhere else (e.g. your desktop) and your sketches folder somewhere else, it's still gonna work.

example-project/project.json
funwithtriangles: You're right in thinking this is how it's stored in the store. I'm a bit scared to have an extra step when saving/loading, it feels like a lot of extra complexity. 

The reason for the IDs in the object is just out of convenience, sometimes you've passed the param around in a function or a component and you need to still get the id later. 

As for having the params nested rather than flat, I see how this would be more readable but I think it could be quite tricky trying to get this to work. Your example would probably be possible but once we have inputs, would you then want those nested inside the params? 😅 What if at some point we want two things to share a node, do we copy the node twice?

I can imagine this getting out of hand, with dependency loops, etc. There is a good reason to keeping the state flat in the store and I think this reason extends to the file format.

However, there is one small thing we can easily do to improve readability: prefix the node IDs with `param_`, `input_`, etc.

cale-bradbury: Good point on the id's, makes a lot of sense as an in-memory object.

I also wasn't sure if you were thinking about shared nodes, I'm into it! Makes it easy to treat an object as a clone vs a separate instance for example.

I don't think we need a prefix for the node, you can get the info you need via the type param.

I do still wonder about the nodeValue in the node, but it's not too big of a deal

funwithtriangles:
Haven't completed all the things I want for this feature but it's good enough for now, the PR is getting quite large! Going to merge and finish the rest later on

PR #450, opened 2024-09-30
glsl support for sketches

funwithtriangles:
Adds glsl support for sketches. The main change is in `SketchesServer.ts`. Most of the changes you see in the sketch file are just because of linting. I also added a type declaration in the example project so TS knows we're importing a string here (eventually I'm hoping we can have these types automatically as an npm package)

### DIFF
--- a/example-project/sketches/hsl/index.ts
+++ b/example-project/sketches/hsl/index.ts
@@ -1,16 +1,16 @@
-import { EffectPass, SavePass, CopyPass, Pass, ShaderPass } from 'postprocessing'
-import { Uniform, Texture, Scene, Camera, WebGLRenderer, WebGLRenderTarget, ShaderMaterial, Vector3 } from 'three';
+import { Pass, ShaderPass } from 'postprocessing'
+import { Uniform, Texture, ShaderMaterial, Vector3 } from 'three'
 
-import fragmentShader from './hsl.js';
+import fragmentShader from './hsl.glsl'
 
 /**
  * A basic post proicessing effect to adjust the hue, saturation, and lightness of the scene.
  * While this could be created with existing postprocessing passes, this example demonstrates how to create a custom effect.
  */
 export default class HSL {
-  shader: ShaderMaterial;
-  pass: ShaderPass;
-  passes:Pass[]
+  shader: ShaderMaterial
+  pass: ShaderPass
+  passes: Pass[]
 
   /**
    * Gets the passes for this effect.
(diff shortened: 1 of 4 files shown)


src/main/SketchesServer.ts
```diff
@@ -13,26 +13,6 @@ const HOST = process.platform.startsWith('win') ? 'localhost' : '0.0.0.0'
...
-  'mp3',
-  'mp4',
-  'ogg',
-  'wav',
-]
-
-const loaderFileExtensions: { [key: string]: 'file' } = {}
-fileExtensions.forEach((ext) => {
-  loaderFileExtensions[`.${ext}`] = 'file'
-})
```
funwithtriangles, line 34: Removed this way of constructing the config option as ultimately I think it's just not that readable.

src/main/SketchesServer.ts
```diff
@@ -87,7 +67,25 @@ export class SketchesServer extends EventEmitter {
...
+        '.jpeg': 'file',
+        '.gif': 'file',
+        '.svg': 'file',
+        '.mp3': 'file',
+        '.mp4': 'file',
+        '.ogg': 'file',
+        '.wav': 'file',
+        // text: loaded into sketch as string
+        '.glsl': 'text',
+        '.isf': 'text',
```
funwithtriangles, line 87: Here's the magic!

example-project/sketches/hsl/index.ts
```diff
@@ -1,16 +1,16 @@
...
+import fragmentShader from './hsl.glsl'
 
 /**
  * A basic post proicessing effect to adjust the hue, saturation, and lightness of the scene.
  * While this could be created with existing postprocessing passes, this example demonstrates how to create a custom effect.
  */
 export default class HSL {
-  shader: ShaderMaterial;
-  pass: ShaderPass;
-  passes:Pass[]
```
cale-bradbury, line 13: haha, oops, I'll get that auto format on save enabled 😅

src/main/SketchesServer.ts
cale-bradbury: This is possibly the way to go though, in the future let sketches/plugins be added to the list.

Either that or find a way to let any file be read as text or imported as a path. Now that I know it's esbuild and not vite, I can bark up the right trees for that

cale-bradbury (approved):
Thanks for this, appreciate it in the short term, will need revisiting in the long term, but I'd call it super low priority

PR #451, opened 2024-10-08
Nodes are selectible

cale-bradbury:
Doesn't do much yet but draw a border,

### DIFF
--- a/src/engine/store/engineStore.ts
+++ b/src/engine/store/engineStore.ts
@@ -9,7 +9,6 @@ import { createDeleteSketch } from './actionCreators/deleteSketch'
 import { createSetSketchModuleItem } from './actionCreators/setSketchModuleItem'
 import { createDeleteSketchModule } from './actionCreators/deleteSketchModule'
 import { createUpdateSketchParams } from './actionCreators/updateSketchParams'
-import { createUpdateSelectedNode } from './actionCreators/updateSelectedNode'
 import { createUpdateNodeValue } from './actionCreators/updateNodeValue'
 import { createReset } from './actionCreators/reset'
 import { createLoadProject } from './actionCreators/loadProject'
@@ -25,7 +24,6 @@ export const createEngineStore = () =>
           updateSketchParams: createUpdateSketchParams(set),
           setSketchModuleItem: createSetSketchModuleItem(set),
           updateNodeValue: createUpdateNodeValue(set),
-          updateSelectedNode: createUpdateSelectedNode(set),
           deleteSketch: createDeleteSketch(set),
           deleteSketchModule: createDeleteSketchModule(set),
           reset: createReset(set),
--- a/src/engine/store/engineStore.ts
+++ b/src/engine/store/engineStore.ts
@@ -5,11 +5,12 @@ import { immer } from 'zustand/middleware/immer'
 import type {} from '@redux-devtools/extension' // required for devtools typing
 import { EngineStateWithActions } from './types'
 import { createAddSketch } from './actionCreators/addSketch'
-import { createSetSketchModuleItem } from './actionCreators/setSketchModuleItem'
-import { createUpdateNodeValue } from './actionCreators/updateNodeValue'
-import { createDeleteSketchModule } from './actionCreators/deleteSketchModule'
 import { createDeleteSketch } from './actionCreators/deleteSketch'
+import { createSetSketchModuleItem } from './actionCreators/setSketchModuleItem'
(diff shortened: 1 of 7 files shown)


src/renderer/components/SketchParams/SketchParams.tsx
```diff
@@ -30,8 +32,11 @@ const getInputElement = (valueType: NodeTypes, id: string) => {
 }
 
 const ParamItem = ({ param: { key, title, id, valueType } }: ParamProps) => {
+  //const onSelectNode = useOnSelectNode(id)
+  const onSelectNode = useEngineStore((state) => state.updateSelectedNode).bind(null, id)
+  const selected = useEngineStore((state) => state.selectedNode)
   return (
-    <NodeControl key={key}>
+    <NodeControl key={key} onClick={onSelectNode} isActive={id === selected}>
```
cale-bradbury: I was originally following the flow of value updates, going through 2 classes in `src/renderer/components/hooks/` (see below)

It appears that I can just hit the state directly via useEngineStore, do we actually want/need these extra files? Do we want to simplify the node value updating too?

funwithtriangles: So my philosophy is to try and reduce the complexity of components as much as possible by creating hooks. The components just get their functionality from hooks that are consuming state. This means the components don't really need to care about what's going on behind the scenes. We could even swap out the whole state management system and the components wouldn't ever need updating.

However I do realise there's a bit too much boilerplate here. We don't need `useUpdateSelectNode`, that could just be implemented inside of `useOnSelectNode`.

src/renderer/components/SketchParams/SketchParams.tsx
```diff
@@ -30,8 +32,11 @@ const getInputElement = (valueType: NodeTypes, id: string) => {
 }
 
 const ParamItem = ({ param: { key, title, id, valueType } }: ParamProps) => {
+  //const onSelectNode = useOnSelectNode(id)
+  const onSelectNode = useEngineStore((state) => state.updateSelectedNode).bind(null, id)
```
funwithtriangles: This use of `.bind` comes with a small performance cost. `.bind` creates a function every time it is called. So whenever `ParamItem` is called, `onSelectNode` will be a new value every time. This gets passed down as a prop and `NodeControl` will need re-rendering every time, because it will evaluate it as a new update to the component. It's probably minuscule in this example, but over time these sorts of things can build up and create little frame skips etc. 

This is the reason why we use something like `useCallback` because it takes a dependency array. Only if the items of the array change will it create a new function. And because of all that annoying stuff, I create a hook to hide it all away from the component 😉

src/engine/store/engineStore.ts
```diff
@@ -24,6 +25,7 @@ export const createEngineStore = () =>
           updateSketchParams: createUpdateSketchParams(set),
           setSketchModuleItem: createSetSketchModuleItem(set),
           updateNodeValue: createUpdateNodeValue(set),
+          updateSelectedNode: createUpdateSelectedNode(set),
```
funwithtriangles: I'd rather this functionality live inside of `appStore` instead of `engineStore`. The `engineStore` should purely be about what's going on inside the scene, nothing UI related. Think about a web project that's running a Hedron scene. It's going to care about nodes and inputs, etc, but it doesn't care about what node or scene the user has selected. By keeping it all separate I'm hoping `hedronEngine` will stay quite slim as a package for web projects. 

If we were following the naming convention for `appStore`, it would be something like:

- `activeNodeId`
- `setActiveNodeId`

However just to complicate things further, i'm wondering if we want this to be an extra level deep. Something like `sketchActiveNodeIds`, which would be an object of sketch IDs mapped to the node ID that is currently active (or `null`/`undefined`). So when switching between sketches we don't lose the current open node for that sketch.

cale-bradbury: good calls, I'll work your feedback in and ping ya when it's ready! thanks!

cale-bradbury: Okay, think it's good now, was a pretty easy change to make and seems to work well for me.

Do we want to have the param deselect if you click it a second time? I'm leaning towards no, as you might be tweaking the value manually, would become a pain to wait and see if you are interacting with a sub-element of it, especially if that's something plugins can have control over. Thinking we just slap a close button on the param editing popup

On to getting familiar with storybook and adding something to make use of this selected state, I'll do that on a separate branch

funwithtriangles:
Works great, thanks for the changes!

PR #455, opened 2024-10-12
Relative and absolute sketches path in project file

funwithtriangles:
- sketches dir is now saved in the project
- project json is now split into `engine` and `app` data
- sketches dir in project can either be an absolute path or a path relative to the parent of the JSON file.
- When the sketches dir is some ancestor of this parent, it is automatically saved as a relative path. This means we can happily send project folders around without having to relocate the sketches dir.
- If the sketches folder doesn't share the same ancestry, we just keep it as an absolute path

### DIFF
--- a/example-project/project.json
+++ b/example-project/project.json
@@ -1,8 +1,5 @@
 {
   "version": 0,
-  "app": {
-    "sketchesDir": "./sketches"
-  },
   "engine": {
     "sketches": {
       "c64ceed4673": {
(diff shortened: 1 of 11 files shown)


funwithtriangles:
@cale-bradbury would be interested if this is all working as expected in windows

cale-bradbury:
> @cale-bradbury would be interested if this is all working as expected in windows

Looks good, project loads with sketches regardless of if the json is in a parent folder or not

PR #456, opened 2024-10-13
Recent projects list

funwithtriangles:
Screenshot 2024-10-13 at 14 30 18

- Displays a recent projects list on the intro screen
- Details for each project shown (num sketches, num scenes, file path, last saved time)
- This data is stored in `appStore` and stored across sessions using zustand's persist middleware (uses `window.localStorage`)


src/renderer/components/Intro/Intro.tsx
```diff
@@ -2,22 +2,74 @@ import c from './Intro.module.css'
...
         <PanelHeader>Welcome back</PanelHeader>
         <PanelBody>Choose your sketches folder or open a project to get started</PanelBody>
         <PanelActions>
           <Button onClick={handleSketchesDialog} iconName="folder_open">
             Select Sketches Folder
           </Button>
-          <Button type="secondary" onClick={handleLoadProjectDialog} iconName="file_open">
-            Open Project
+          <Button type="secondary" onClick={() => handleLoadProjectDialog()} iconName="file_open">
+            Find Project File
```
funwithtriangles, line 61: We could use `useCallback` here to help with performance but I don't think it's necessary for the intro screen

src/renderer/components/Intro/Intro.tsx
```diff
@@ -2,22 +2,74 @@ import c from './Intro.module.css'
...
           </Button>
         </PanelActions>
       </Panel>
+      <h2 className={c.projectsTitle}>Recent Projects</h2>
+      <CardList>
+        {saveList.map((item) => (
+          <ErrorBoundary key={item.path} fallbackRender={() => ''}>
+            <RecentProjectItem item={item} />
+          </ErrorBoundary>
```
funwithtriangles, line 70: Need to have an `ErrorBoundary` here because `date-fns` will throw an exception if there is some bad formatting on dates, crashing the app.

PR #466, opened 2024-11-04
Monorepo

funwithtriangles:
I've been to config hell and back! We now have a monorepo, with three packages:

- `@hedron/engine` (will eventually be on npm)
- `@hedron/desktop` (depends on `@hedron/engine`)
- `@hedron/example-project` (still all sorts of type errors here but we can worry about that later)

For some boring technical reason we're back to using `yarn`. You can do everything from the root of the project, see the readme file. It's important that `@hedron/engine` is built before you start developing, otherwise you'll see a bunch of errors. `yarn dev` should do everything for you.

Let me know if this is working for you @cale-bradbury. I'm also fully aware how much this is going to mess any PRs you might have coming in. Not sure what's going to be easier to merge first, although I feel like there's less chance of changes getting lost if we merge this first and then you'll have to manually move your changed files into this new system. 

**Last fixes**
- [x] add dependencies to engine package.json
- [x] remove any drops from desktop package.json


cale-bradbury:
sorry for the late reply, it's working on windows after a `yarn` / `yarn dev`

PR #471, opened 2024-12-06
bad sketches dont block run

cale-bradbury:
I had a sketch that was now failing due to the removal of `THREE.GLTFLoader`, that said, there could be any number of things that break a sketch.

This change continues with the initialization so you can at least load sketches that did init properly.

The sketch that failed does not appear in the list of sketches you can add.

This might leave an issue when opening a project that wants to use an uninitialized sketch due to bad changes, but I doubt it would break any worse than failing to init the three scene

### DIFF
--- a/packages/engine/src/HedronEngine.ts
+++ b/packages/engine/src/HedronEngine.ts
@@ -36,7 +36,11 @@ export class HedronEngine {
 
   public async initiateSketchModules(moduleIds: string[]) {
     for (const moduleId of moduleIds) {
-      await this.addSketchModule(moduleId)
+      try {
+        await this.addSketchModule(moduleId)
+      } catch (error) {
+        console.error('Init error in: ' + moduleId, error)
+      }
     }
 
     this.store.setState({ isSketchModulesReady: true })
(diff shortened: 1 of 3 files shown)


funwithtriangles:
Thanks for this, clearly an important topic that needed to be addressed. I was looking at the code and realised that it probably makes more sense to handle this at the lowest level possible, so that we're catching this when sketches are refreshed on code changes, not just when loaded on init.

I ended up in a little rabbit hole but I think the result is quite nice. We now explicitly return a `Result` object when importing sketches, with a `success` boolean flag. Functions higher up can decide what to do with that (e.g. later we'll want to do something with the UI to explain an error).

packages/engine/src/HedronEngine.ts
```diff
@@ -42,21 +43,34 @@ export class HedronEngine {
...
   }
 
-  public async reimportSketchModuleAndReloadSketches(moduleId: string) {
-    const moduleItem = await this.addSketchModule(moduleId)
+  public async reimportSketchModuleAndReloadSketches(moduleId: string): Promise<void> {
+    const result = await this.addSketchModule(moduleId)
+
+    if (!result.success) {
+      return
+    }
```
funwithtriangles, line 71: Because we're now checking for this issue lower down, other functions are affected. Here we're just not doing the reimport stuff if the module failed to load.

cale-bradbury: yeah, this looks really nice, feel free to merge :)

PR #477, opened 2024-12-18
vector3 param type

funwithtriangles:
This introduces a new type of param to the core, `vector3`. It has a `childNodeIds` property which points to three more param nodes with the `number` value type. It feels like a nicer way to do it so we can still link inputs to these child nodes as if they were any other number node, rather than dealing with actual array values. 

At the last step of the engine, the array gets created from the three child nodes and passed as a param to the sketch. I think we can merge this sooner rather than later, in future PRs, I plan to:

- Improve the component, so the three sliders popout. This will allow for individual inputs to be linked.
- Add `vector2`, `rgb`, `hsl` types, each with their own special component (e.g. color picker)


funwithtriangles:
@cale-bradbury what do ya think? Does this approach make sense to you? We'll have special slider components for each type (e.g. a color picker, XY field) but always with a "popout" option which exposes the underlying number nodes where users can assign inputs as always.

packages/engine/src/store/selectors/getSketchParamValues.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  sketch.paramIds.forEach((id) => {
+    const { key, valueType } = nodes[id]
+
+    let value
+
+    if (valueType === 'vector3') {
+      // Return an array of values for vector3
+      const childNodeIds = nodes[id].childNodeIds
+      value = childNodeIds.map((childNodeId) => nodeValues[childNodeId])
```
funwithtriangles, line 18: We're getting all child nodes and creating an array to be handed to the sketch. Is that how users will want it? We could also hand it back as an object with `x`, `y` and `z` properties (or `r`, `g`, `b` for colors...). Part of me thinks an array is simplest and the user can deal with it how they please.

packages/example-project/sketches/solid/index.ts
```diff
@@ -39,6 +40,8 @@ export default class Solid {
     this.root.rotation.y += params.rotSpeedY * baseSpeed * deltaFrame
     this.root.rotation.z += params.rotSpeedZ * baseSpeed * deltaFrame
 
+    this.root.position.set(...params.position)
```
funwithtriangles, line 43: Here the array format works perfectly for three js, having it as object with `x`, `y` and `z` properties would just be more work, but I can imagine other times it would be nicer to have it the other way around. However for most cases, including color, three.js plays nicely with plain array so maybe we stick with the simplest approach.

PR #491, opened 2025-01-28
Basic Midi Support

cale-bradbury:
It's pretty clunky, and not very pretty, but a first pass at getting midi/midi learn working.

Confirmed to work with ccs, probably works with buttons

Went in with generic events in mind, figure most things could go through the same functions.

It's not at all set up for it, since midi is treated with quite a bit of privilege here, but still percolating on sketches-as-plugins idea, inputs coming from sketches, would need a good bit of architecture to say enable midi to have a midi learn popup where-as other inputs might need their own special thing, lfos with their additional per-param settings, etc. Don't think it needs to be solved at the moment, just sleepy spouting thought-trains choo-choo

Edit: Forgot about midival while writing this, could look into moving the code over in the future, it didn't take much code to get this working with vanilla js though. 

Also built off the clock branch so targeting that, if that branch is unlikely to merge soon, I can cherry pick the commits while targeting alpha

### DIFF
--- a/packages/desktop/src/renderer/components/MidiSetting/MidiSetting.tsx
+++ b/packages/desktop/src/renderer/components/MidiSetting/MidiSetting.tsx
@@ -1,0 +1,39 @@
+import { NodeTypes, ParamWithInfo } from '@hedron/engine'
+import { useInputsWithNode } from '@components/hooks/useInput'
+import { engine } from '@renderer/engine'
+
+interface ParamMidiSetting {
+  param: ParamWithInfo
+}
+
+export const MidiSetting = ({ param }: ParamMidiSetting) => {
+  const inputs = useInputsWithNode(param.id)
+  if (param.valueType !== NodeTypes.Number) {
+    return
+  }
+
+  async function useMidiLearn() {
+    await engine.midiLearn(param.id)
+  }
+
+  function removeMidi(id: string) {
+    return () => {
+      engine.deleteInputParam(id, param.id)
+    }
+  }
+
+  return (
+    <>
+      {param.id} - {param.valueType}
+      {inputs.map((input) => (
+        <div key={input.id}>
+          {input.id} - {input.type}
+          <button onClick={removeMidi(input.id)}>Remove</button>
+        </div>
+      ))}
+      <div>
+        <button onClick={useMidiLearn}>Midi Learn</button>
+      </div>
+    </>
+  )
+}
--- a/packages/desktop/src/renderer/components/hooks/useInput.ts
+++ b/packages/desktop/src/renderer/components/hooks/useInput.ts
@@ -1,0 +1,11 @@
+import { useEngineStore } from '@renderer/engine'
+
+export const useInput = (id: string) => {
+  const inputs = useEngineStore((state) => state.inputs)
+  return inputs[id]
+}
+
+export const useInputsWithNode = (nodeId: string) => {
+  const inputs = useEngineStore((state) => state.inputs)
+  return Object.values(inputs).filter((input) => input.targetNodeIds.includes(nodeId))
+}
--- a/packages/engine/package.json
+++ b/packages/engine/package.json
@@ -28,6 +28,7 @@
     "three": "*"
   },
   "dependencies": {
+    "@hedron/midi": "^1.0.0-alpha.0",
     "lodash.debounce": "^4.0.8",
     "uid": "^2.0.2",
     "zustand": "^5.0.2"
--- a/packages/engine/src/HedronEngine/HedronEngine.ts
+++ b/packages/engine/src/HedronEngine/HedronEngine.ts
@@ -1,3 +1,4 @@
+import { Midi, MIDIEvent } from '@hedron/midi'
 import { listenToStore } from './storeListener'
 import { Result } from './types'
 import { importSketchModule } from './importSketchModule'
@@ -5,7 +6,7 @@ import { stripForSave } from '@utils/stripForSave'
 import { Renderer } from '@world/Renderer'
 import { SketchManager } from '@world/SketchManager'
 import { createDebugScene } from '@world/debugScene'
-import { EngineData, SketchModuleItem } from '@store/types'
+import { EngineData, Input, SketchModuleItem } from '@store/types'
 import { getSketchesOfModuleId } from '@store/selectors/getSketchesOfModuleId'
 import { createEngineStore, EngineStore } from '@store/engineStore'
 import { getSketchParamValues } from '@store/selectors/getSketchParamValues'
@@ -15,11 +16,49 @@ export class HedronEngine {
   private store: EngineStore
   private sketchesUrl: string | null = null
   private sketchManager: SketchManager
+  private midi: Midi
 
   constructor() {
     this.store = createEngineStore()
     this.sketchManager = new SketchManager()
     this.renderer = new Renderer()
+    this.midi = new Midi()
+    this.midi.onMidiMessage.add((event) => {
+      const id = `${event.device.name} - ${event.message.data?.[1]}`.replace(/ /g, '_')
+      this.store.getState().updateInputValues(id, (event.message.data?.[2] || 0) / 127)
+    }, this)
+  }
+
+  public async awaitNextMidiMessage(): Promise<MIDIEvent | null> {
+    // null response would be canceling the midi learn, but that is not hooked up yet as there is no GUI
+    return new Promise((resolve) => {
+      const listener = (event: MIDIEvent) => {
+        this.midi.onMidiMessage.remove(listener)
+        resolve(event)
+      }
+      this.midi.onMidiMessage.add(listener)
+    })
+  }
+
+  public async midiLearn(paramId: string): Promise<Input> {
+    const event = await this.awaitNextMidiMessage()
+    if (!event) {
+      throw new Error('No MIDI event received')
+    }
+
+    const id = `${event.device.name} - ${event.message.data?.[1]}`.replace(/ /g, '_')
+    let input = this.store.getState().inputs[id]
+    if (!input) {
+      input = {
+        id,
+        type: 'midi',
+        targetNodeIds: [paramId],
+      }
+      this.store.getState().addInput(id, input)
+      return input
+    }
+    this.store.getState().addInputParam(id, paramId)
+    return input
   }
 
   public setSketchesUrl(sketchesUrl: string) {
@@ -100,6 +139,10 @@ export class HedronEngine {
 
   public getSaveData(): EngineData {
     return stripForSave(this.store.getState())
+  }
+
+  public deleteInputParam(inputId: string, nodeId: string) {
+    this.store.getState().deleteInputParam(inputId, nodeId)
   }
 
   run() {
--- a/packages/engine/src/store/actionCreators/createAddInput.ts
+++ b/packages/engine/src/store/actionCreators/createAddInput.ts
@@ -1,0 +1,31 @@
+import { Input, SetterCreator } from '@store/types'
+
+export const createAddInput: SetterCreator<'addInput'> =
+  (setState) => (inputId: string, value: Input) => {
+    setState((state) => {
+      if (state.inputs[inputId]) {
+        return
+      }
+      state.inputs[inputId] = value
+    })
+  }
+
+export const createAddInputParam: SetterCreator<'addInputParam'> =
+  (setState) => (inputId: string, nodeId: string) => {
+    setState((state) => {
+      const input = state.inputs[inputId]
+      if (input) {
+        input.targetNodeIds.push(nodeId)
+      }
+    })
+  }
+
+export const createDeleteInputParam: SetterCreator<'deleteInputParam'> =
+  (setState) => (inputId: string, nodeId: string) => {
+    setState((state) => {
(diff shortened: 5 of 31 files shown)


packages/desktop/src/renderer/components/MidiSetting/MidiSetting.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { engine } from '@renderer/engine'
+
+interface ParamMidiSetting {
+  param: ParamWithInfo
+}
+
+export const MidiSetting = ({ param }: ParamMidiSetting) => {
+  const inputs = useInputsWithNode(param.id)
+  if (param.valueType !== NodeTypes.Number) {
+    return
```
cale-bradbury, line 12: could likely add nice support for other types, just not in the first pass

packages/desktop/src/renderer/components/hooks/useInput.ts
```diff
@@ -0,0 +1,11 @@
+import { useEngineStore } from '@renderer/engine'
+
+export const useInput = (id: string) => {
+  const inputs = useEngineStore((state) => state.inputs)
+  return inputs[id]
+}
+
+export const useInputsWithNode = (nodeId: string) => {
```
cale-bradbury, line 8: if you want I can bump these into seperate files, just feels a tad bloaty

packages/engine/src/HedronEngine/HedronEngine.ts
```diff
@@ -1,11 +1,12 @@
+import { Midi, MIDIEvent } from '@hedron/midi'
```
cale-bradbury, line 1: for some reason this isn't recognised in my vscode, despite being a dependency for engine, maybe it needs to be bumped up to the root?

packages/engine/src/HedronEngine/HedronEngine.ts
```diff
@@ -15,11 +16,49 @@ export class HedronEngine {
   private store: EngineStore
   private sketchesUrl: string | null = null
   private sketchManager: SketchManager
+  private midi: Midi
```
cale-bradbury, line 19: Likely want to move this out, and into an InputManager class

packages/engine/src/HedronEngine/HedronEngine.ts
```diff
@@ -15,11 +16,49 @@ export class HedronEngine {
...
 
   constructor() {
     this.store = createEngineStore()
     this.sketchManager = new SketchManager()
     this.renderer = new Renderer()
+    this.midi = new Midi()
+    this.midi.onMidiMessage.add((event) => {
+      const id = `${event.device.name} - ${event.message.data?.[1]}`.replace(/ /g, '_')
```
cale-bradbury, line 27: This form of ID feels bad, just something quick for testing. Likely want it to also be generated as some kind of UUID on the MIDIEvent class.

Noticed there is a device.id, but unsure if that remains consistent between boots/usb port used. Though being able to use 2 of the same controller and have them mapped automattically, it's a feature! >.<

packages/engine/src/store/actionCreators/createUpdateInputValues.ts
```diff
@@ -0,0 +1,16 @@
...
+  (setState) => (inputId: string, value: NodeValue) => {
+    setState((state) => {
+      const input = state.inputs[inputId]
+      if (!input) {
+        //throw new Error(`Input with id ${inputId} not found.`)
+        return
+      }
+
+      input.targetNodeIds.forEach((nodeId) => {
+        state.nodeValues[nodeId] = value
```
cale-bradbury, line 13: In the future, we may want some kind of value merging for when multiple inputs target the same value (additive, etc) Would likely want a whole UI for that though, feels pretty far off in terms of needs

packages/engine/src/store/types.ts
```diff
@@ -135,10 +135,19 @@ export type SketchModules = { [key: string]: SketchModuleItem }
 
 export type EnumOption = { value: string; label: string }
 
+export interface Input {
+  id: string
+  type: 'midi' | 'gamepad' | string
```
cale-bradbury, line 140: haven't really thought type through, might want to use the event types from the midi package, unused atm, could delete

packages/desktop/src/renderer/components/MidiSetting/MidiSetting.tsx
funwithtriangles: yep makes sense to keep it simple for now!

packages/desktop/src/renderer/components/MidiSetting/MidiSetting.tsx
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  async function useMidiLearn() {
```
funwithtriangles, line 15: Just a nitpick, but anything starting with `use` makes me think of a React hook, so maybe we should call this something else? `startMidiLearn`?

packages/desktop/src/renderer/components/MidiSetting/MidiSetting.tsx
```diff
@@ -0,0 +1,39 @@
...
+    await engine.midiLearn(param.id)
```
funwithtriangles, line 16: and then maybe this should also be `startMidiLearn`?

packages/desktop/src/renderer/components/hooks/useInput.ts
funwithtriangles: totally fine with it in one file!

packages/desktop/src/renderer/components/hooks/useInput.ts
```diff
@@ -0,0 +1,11 @@
...
+  const inputs = useEngineStore((state) => state.inputs)
+  return Object.values(inputs).filter((input) => input.targetNodeIds.includes(nodeId))
```
funwithtriangles, line 10: Probably worth using `useMemo` here, to save a bit of computation

packages/engine/src/HedronEngine/HedronEngine.ts
funwithtriangles: Working fine for me! Do you have a `/dist` file in the midi package dir? That is definitely needed (Should happen whenever you run `yarn dev` or `yarn build`)

packages/engine/package.json
```diff
@@ -28,6 +28,7 @@
     "three": "*"
   },
   "dependencies": {
+    "@hedron/midi": "^1.0.0-alpha.0",
```
funwithtriangles, line 31: This is fine for now, but I think eventually we'll want to pass these dependencies into the engine on runtime as plugins to keep the engine file size as small as possible for web projects.

packages/engine/src/HedronEngine/HedronEngine.ts
funwithtriangles: I was able to get it to error in VSCode by deleting the `/dist` folder, it took a bit of fiddling to get it to disappear after I put the folder back. Had to close the tab, restart TS server, open the tab.

packages/engine/src/HedronEngine/HedronEngine.ts
funwithtriangles: Yeah I think it makes sense to get this separated before merging, just to keep things tidy and prevent merge conflicts if we're both working on the engine, including all the new public methods below.

packages/engine/src/HedronEngine/HedronEngine.ts
funwithtriangles: I think we should ignore the device ID entirely and just use the MIDI channel (and note) to give the input its ID. Then it's just up to the user to make sure their devices are on separate channels, as per the MIDI spec. I think the device ID is useful in other ways but probably not on a per-message basis.

packages/engine/src/HedronEngine/HedronEngine.ts
```diff
@@ -15,11 +16,49 @@ export class HedronEngine {
...
+  public async awaitNextMidiMessage(): Promise<MIDIEvent | null> {
+    // null response would be canceling the midi learn, but that is not hooked up yet as there is no GUI
+    return new Promise((resolve) => {
+      const listener = (event: MIDIEvent) => {
+        this.midi.onMidiMessage.remove(listener)
+        resolve(event)
+      }
+      this.midi.onMidiMessage.add(listener)
+    })
+  }
```
funwithtriangles, line 41: I love this simple approach! I would have gone about this in a far more convoluted way 😅 How would the cancellation work? Would `awaitNextMidiMessage` have an extra listener for some cancel event?

packages/engine/src/store/actionCreators/createAddInput.ts
```diff
@@ -0,0 +1,31 @@
...
+  (setState) => (inputId: string, value: Input) => {
+    setState((state) => {
+      if (state.inputs[inputId]) {
+        return
+      }
+      state.inputs[inputId] = value
+    })
+  }
+
+export const createAddInputParam: SetterCreator<'addInputParam'> =
```
funwithtriangles, line 13: I wasn't sure what this was doing until I got to this file. Perhaps the name should be more explicit? `addInputTargetNodeId`?

packages/engine/src/store/actionCreators/createAddInput.ts
```diff
@@ -0,0 +1,31 @@
...
+  (setState) => (inputId: string, nodeId: string) => {
+    setState((state) => {
+      const input = state.inputs[inputId]
+      if (input) {
+        input.targetNodeIds.push(nodeId)
+      }
+    })
+  }
+
+export const createDeleteInputParam: SetterCreator<'deleteInputParam'> =
```
funwithtriangles, line 23: Same sort of thing as above, maybe `removeInputParamTargetNodeId`

packages/engine/src/store/actionCreators/createUpdateInputValues.ts
funwithtriangles: hmm! never thought of this sort of thing, but yeah definitely a bit further down the line 😛

packages/engine/src/store/actionCreators/createUpdateInputValues.ts
```diff
@@ -0,0 +1,16 @@
+import { NodeValue, SetterCreator } from '@store/types'
+
+export const createUpdateInputValues: SetterCreator<'updateInputValues'> =
```
funwithtriangles, line 3: I'm not totally sure if this should even be a store action, as we're mutating state. I realise this exactly how i'd done it for updating values with `updateNodeValue` but now I'm having a rethink. No need to change in this PR, more of a note for myself! I will address this later.

packages/engine/src/store/types.ts
funwithtriangles: It's fine for now, I think this is just going to naturally evolve as we add more inputs and realise what we do and don't need! I think for now though you can just keep it as `midi` with no extras and we just add to that.

packages/midi-test-app/src/App.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  function onMidiMessage({ device, message }: MIDIEvent) {
+    if (!message.data) {
+      console.error('No data in MIDI message:', message)
+      return
+    }
+    const status = message.data[0]
+    const statusType = midi.getMidiMessageType(status)
+    if (statusType === MidiMessageType.Unknown) {
+      return // my launch pad was sending unknown non-stop, would imagine some other devices do as well
```
funwithtriangles, line 22: Curious to know what those messages actually are...

funwithtriangles (changes requested):
This is really great @cale-bradbury, the general approach makes a lot of sense and I think we can get this merged soon. I've commented where appropriate, but just to summarise the two main changes I think you should implement before merging:

1. Move all the MIDI stuff into a single class inside the Hedron engine, so that `HedronEngine.ts` only exposes a single `midi` property.
2. Don't use the device ID to key the inputs, we should be using the channel instead.

Let me know what you think! And thanks! 🙌

packages/midi/src/index.ts
```diff
@@ -0,0 +1,170 @@
...
+   */
+  private clearMidiEventListeners = (): void => {
+    this.devices.forEach((device) => {
+      const listener = this.eventListeners.get(device)
+      if (listener) {
+        device.removeEventListener('midimessage', listener)
+        this.eventListeners.delete(device)
+      }
+    })
+  }
```
funwithtriangles, line 73: We probably want this method to be public. Or wrapped in some `destroy` sort of method. e.g. needed in React when a component unmounts.

funwithtriangles: I have used a stripped down version of your Midi class in Clock, but `clearMidiEventListeners` is now async and has some extra promise logic. The reason behind this, is that if someone created the class and then *immediately* tried to clear the listeners, it doesn't work because the listeners are not created yet (as the device finding is async). This might seem like a weird edge case but actually in React when you're in dev mode, `useEffect` is called twice in succession as a way to force you to be clean with your code.

Here's how I solved it:
https://github.com/nudibranchrecords/hedron/pull/492/files#diff-b6e106463b54d5e5622bd46ba9118fd94fa768262c379eebb4df919003409fea

funwithtriangles:
@cale-bradbury this got automatically closed because I merged in the clock branch and deleted it. I think you just need to repoint it to `alpha` now?

PR #498, opened 2025-02-21
[WIP] Clock display

funwithtriangles:
Displays BPM with controls under the viewer. Still need to:

- [x] Tidy up how the BPM edit works
- [x] nice hover/focus state for BPM to show its editable
- [x] maybe use input instead of contenteditable
- [x] submit on ENTER
- [x] all buttons as a row on bottom?
- [x] pause / play as toggle, stop as separate?


packages/desktop/src/renderer/components/GlobalClock/GlobalClock.tsx
```diff
@@ -0,0 +1,38 @@
+import { useEffect, useState } from 'react'
+import { ClockDisplay } from '@components/core/ClockDisplay/ClockDisplay'
+import { clock } from '@renderer/engine'
+
+const onBpmEdit = (bpm: number) => {
+  clock.bpm = bpm
+}
+
+export const GlobalClock = () => {
```
cale-bradbury, line 9: I kind of ran into this with my midi work, Feel like it would be good if a packages ui etc lived in the package so all the code stays contained, easy to upgrade/switch package versions.

I tried making a quick plugin system, but feel into an issue with hooks, maybe you might have a better idea for how to get the tsx living in the subpackage. I made a pr into my working midi branch showing the changes, I'll highlight the code of note https://github.com/cale-bradbury/hedron/pull/9

funwithtriangles: I agree that generally we should do this with the subpackages, but for now I'm gonna keep this one in the main package. Maybe I'll clean it up later if it feels right!

PR #499, opened 2025-03-21
Midi Plugin

cale-bradbury:
image

plug n play midi binding to number params


packages/engine/src/plugin/Plugin.ts
```diff
@@ -0,0 +1,25 @@
...
+     * The description of the plugin.
+     */
+    description: string;
+
+    /**
+     * Gets the selected parameter view.
+     * @param param - The parameter with info.
+     * @returns The JSX element or undefined if this plugin does not have a visual element.
+     */
+    getSelectedParamView: (param: ParamWithInfo) => JSX.Element | undefined;
```
cale-bradbury: I'm thinking starting with 2 optional functions for plugins, one for the selected param display, and one as a 'main' display (ie clock under the render view)

packages/midi/src/MidiSetting.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * Get the inputs that are connected to the node
+ * @param nodeId The node id to get the inputs for
+ * @param useEngineStore useEngineStore hook
+ * @returns The inputs that are connected to the node
+ */
+const useInputsWithNode = (nodeId: string, useEngineStore: UseEngineStore) => {
+  const inputs = useEngineStore((state) => state.inputs)
+  return Object.values(inputs).filter((input) => input.targetNodeIds.includes(nodeId))
```
cale-bradbury: This is grabbing all inputs, might want a better way to register inputs to the plugin, so it can just get the inputs it is meant to fill

packages/engine/src/plugin/Plugin.ts
funwithtriangles: I think eventually we'll want plugins to have types. So an `input` type might have something like `paramView` and a `settingsView`. Perhaps something more general might be a `panel` type that has a `panelView` and also `settingsView`. Not sure yet! For now I'd say just keep it simple and have only one view for inputs, unless you already have an idea for what `main` might be?

funwithtriangles: In terms of showing MIDI devices, do you see that as part of the same plugin? I'd suggest that the MIDI device manager is just a permanent core feature (can still live as a separate package) that all plugins can talk to?

packages/desktop/src/renderer/components/hooks/useInput.ts
```diff
@@ -0,0 +1,11 @@
+import { useEngineStore } from '@renderer/engine'
+
+export const useInput = (id: string) => {
+  const inputs = useEngineStore((state) => state.inputs)
+  return inputs[id]
+}
```
funwithtriangles: Some performance things to consider here. `useEngineStore` will cause a component re-render if it is returning a new object compared to the last render.  Whenever an input changes (e.g. elsewhere in the app an input is added), the `inputs` object will update, causing `useEngineStore` to trigger a re-render. You can fix this by directly selecting inside `useEngineStore`. You don't even need to have this in a separate hook if you don't want to, it's just nice sometimes to keep this abstracted away (especially if it then becomes more complex later on).

```suggestion
export const useInput = (id: string) => {
  return useEngineStore((state) => state.inputs[id])
}
```

packages/desktop/src/renderer/components/hooks/useInput.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export const useInputsWithNode = (nodeId: string) => {
+  const inputs = useEngineStore((state) => state.inputs)
+  return Object.values(inputs).filter((input) => input.targetNodeIds.includes(nodeId))
+}
```
funwithtriangles: Similar issue here, but a bit more complex. Your filtered list of inputs will always return a new array, thus triggering a re-render. `Object.values` and `Array.filter` always generates a new list, and so will always cause a re-render. Zustand provides a hook to fix this, `useShallow`, which performs a shallow equality check on your filtered output to see if it actually changed.

```ts
export const useInputsWithNode = (nodeId: string) => {
  return useEngineStore(
    useShallow((state) => 
      Object.values(state.inputs).filter((input) => input.targetNodeIds.includes(nodeId))
    )
  );
};
```

packages/desktop/src/renderer/components/ActiveSketch/ActiveSketch.tsx
```diff
@@ -18,6 +23,14 @@ export const ActiveSketch = () => {
 
   const selectedParam = useSelectedParam()
 
+  const pluginView = engine.plugins.map((plugin) => (
```
funwithtriangles, line 33: Grabbing values directly from the engine object makes me nervous here, because React has no way of knowing if that object has changed and won't update accordingly. It's fine for now because I think in practice the plugins will always load before this component is mounted and then never change, just worth noting that we'll probably want to clean this up at some point.

packages/desktop/src/renderer/engine.ts
```diff
@@ -18,3 +19,4 @@ export const engineStore = engine.getStore()
 export const useEngineStore = <T>(selector?: (state: EngineStateWithActions) => T) => {
   return useStore(engineStore, selector!)
 }
+engine.registerPlugin(new Midi(engine, useEngineStore))
```
funwithtriangles, line 22: It's so cool to think that we'll only need to make changes here in `@hedron/desktop` to get new functionality! Very tidy

packages/engine/src/HedronEngine/HedronEngine.ts
```diff
@@ -26,6 +29,11 @@ export class HedronEngine {
     this.onFrameEnd = params?.onFrameEnd
   }
 
+  public registerPlugin(plugin: IPlugin) {
+    // plugins isn't actually used anywhere yet, but I assume it will be at some point
```
funwithtriangles: Comment is already out of date 😉

packages/engine/src/HedronEngine/HedronEngine.ts
```diff
@@ -26,6 +29,11 @@ export class HedronEngine {
...
+    this.plugins.push(plugin)
```
funwithtriangles, line 33: Not something we need now, but related to a comment above, we'll probably want to fire some event so that the desktop can update it's own `plugins` state, in case a plugin is registered after initial load of the app. Really don't think we need it anytime soon though!

packages/midi/src/MidiSetting.tsx
funwithtriangles: Hmm we have a duplicate hook inside `/hooks`. We're definitely going to want a separate package at some point soon for useful things like this! See my above comment about how to optimise this hook.

funwithtriangles (approved):
This is cool! It works! I'm happy to merge this soon, just a couple of little points, if you feel like doing them. Otherwise we can merge it and I can tidy stuff up as I work on the LFO stuff anyway. Looks like I can already do it as a plugin 😎

packages/desktop/src/renderer/engine.ts
cale-bradbury: could probably even avoid this with decorators, or having the plugin register itself, not worth worrying about yet though

cale-bradbury:
@funwithtriangles good to merge whenever you would like

edit: seems like it's failing the lint checks, not sure why, it's saying that fields are missing in props validation that are a part of the type :/

funwithtriangles:
Thanks for this! I fixed a few config/linting things to get it merged

PR #500, opened 2025-03-26
UI Core package

funwithtriangles:
Extracts all of the `core` UI components into it's own package, that can then be used by other packages. I had to do some other config adjustments to get things working nicely:

- `@hedron/ui-core` and `@hedron/midi` are imported into `@hedron/desktop` using aliases (setup in the vite config). We are pulling the files directly from source so that hot module reloading works.
- Setup a new eslint rule so that any react components are only ever exporting one thing from a file. This is also needed from vite to allow hot module reloading (aka fast refresh) to work


packages/midi/src/MidiSetting.tsx
```diff
@@ -10,23 +11,6 @@ interface IProps {
...
- */
-export function getMidiSetting(
-  param: ParamWithInfo,
-  engine: HedronEngine,
-  useEngineStore: UseEngineStore,
-  midi: Midi,
-): JSX.Element {
-  return <MidiSetting param={param} engine={engine} useEngineStore={useEngineStore} midi={midi} />
-}
-
```
funwithtriangles, line 29: @cale-bradbury  Had to move this out of `MidiSetting` so we could get hot reloading to work. It means that the whole app doesn't restart when making changes to the plugin components. [More info ](https://github.com/vitejs/vite-plugin-react-swc?tab=readme-ov-file#consistent-components-exports)in case you're interested.

.eslintrc.cjs
```diff
@@ -14,6 +15,7 @@ module.exports = {
     'import/no-named-as-default-member': 0,
     'import/order': 'error',
     'import/no-extraneous-dependencies': 'error',
+    'react-refresh/only-export-components': ['error', { allowConstantExport: true }],
```
funwithtriangles, line 18: We now have an eslint rule enforcing proper setup of react components so that hot reloading will work 
https://github.com/vitejs/vite-plugin-react-swc?tab=readme-ov-file#consistent-components-exports

PR #501, opened 2025-03-30
Text Input & String Param

cale-bradbury:
Adds a text input ui component, and a string param to allow setting a string in a sketch

My sketch is no good yet, but I'll add it to the examples once I've got a good one 

image


packages/desktop/src/main/SketchesServer.ts
```diff
@@ -85,6 +85,7 @@ export class SketchesServer extends EventEmitter {
         // text: loaded into sketch as string
         '.glsl': 'text',
         '.isf': 'text',
+        '.ply': 'text',
```
cale-bradbury: sneaking this in, would be good if there is a way to import anything unknown as text

packages/desktop/src/renderer/components/GlobalDialogs/SketchModulesDialog.tsx
```diff
@@ -69,7 +69,7 @@ export const SketchModulesDialog = ({ closeDialog }: GlobalDialogProps) => {
         <PanelHeader iconName="add_circle" buttonOnClick={closeDialog}>
           Add sketch to scene
         </PanelHeader>
-        <PanelBody>
+        <PanelBody scrollable={true}>
```
cale-bradbury: 😎 wrote enough ts sketches

packages/desktop/src/main/SketchesServer.ts
funwithtriangles: Good point. I feel like there should be a way to do this with esbuild but a quick search didn't show much. Did you try with `'.*': 'text'` ?

packages/desktop/src/renderer/components/GlobalDialogs/SketchModulesDialog.tsx
funwithtriangles: A lot more than me 😆

funwithtriangles:
Hey @cale-bradbury this all looks fine but I'm hesitant to merge it until there's an example sketch to test with. Will keep this open until then!

funwithtriangles:
https://github.com/user-attachments/assets/be9db96d-e9aa-4033-a65a-ec4262dc1b0f

Simple example sketch added. This works great, thanks!

PR #502, opened 2025-03-30
Inputs refactor

funwithtriangles:
Some refactoring to how inputs as plugins work. Overview:

- `@hedron/midi` goes back to being a standalone MIDI helper. It knows nothing about the engine, just handles MIDI devices and messages, etc
- `@hedron/midi-input` is the actual plugin. It has two main parts:
   - `MidiInput` the plugin class. This is handed to the engine with `engine.registerPlugin`. Deals with listening to midi messages and updating the store. Does not know anything about React or any sort of views for the plugin.
   - `MidiInputPanel` is the React component for the input. It just takes two props: `input` and `engine`. I'm hoping we can have this as standardised for all inputs. This view is not registered with the engine. It is registered with `@hedron/desktop` for now just as a property of a `pluginViews` object.
- Listing, adding and removing of inputs are now handled by `@hedron/desktop`, not the plugin itself. The component `SelectedParam` is where that happens, and it grabs `MidiInputPanel` from the `pluginViews` object, handing it the `input` and `engine` props
- The midi input is created first with some default options, then you use midi learn to update the note/channel info. Plugins have a `generateInitialOptions` method to do this.
- In the store, every input now has an `options` object that is custom for whatever that input is doing (e.g. `channel` and `note` info)
- Inside a component, plugins can grab whatever special functionality they need from the engine, as the plugin class is attached to the engine. E.g. doing this for MIDI learn inside of `MidiInputPanel`
- No longer using the specially constructed IDs for the midi input. Instead we listen to events coming in and loop through all inputs and filter based on `options.note` and `options.channel`. This is clearly not very efficient but I think we can solve that with some cacheing system later on. The idea behind this is I want the inputs to be as general as possible with standard UUIDs so we can create them at the level of the app rather than the plugin.

I have quite a lot of tidying up I need to do before merging this, and have a few more plans for restructuring the data a bit. I'm thinking the `inputs` object in the store should be broken into one extra level, with properties for the different inputs (e.g. `midi`, `lfo`, `audio`). It would make things more efficient because inputs like `audio` and `lfo` are going to be looping through all of the inputs anyway, so better to keep the list specific I think.


packages/desktop/src/renderer/components/SelectedParam/SelectedParam.tsx
```diff
@@ -0,0 +1,67 @@
...
+      options: engine.plugins['midi-input'].generateInitialOptions(),
+    }
+
+    addInput(input)
+  }, [addInput, selectedParam.id])
+
+  const inputs = useInputsWithNode(selectedParam.id)
+
+  // TODO: "midi-input" should not be hardcoded here
+  const PluginView = currentInput && pluginViews['midi-input'].inputPanel
```
funwithtriangles: We grab the view for inputs like this

packages/desktop/src/renderer/engine.ts
```diff
@@ -19,4 +19,10 @@ export const engineStore = engine.getStore()
...
   return useStore(engineStore, selector!)
 }
-engine.registerPlugin(new Midi(engine, useEngineStore))
+engine.registerPlugin(new MidiInput(engine))
+
+export const pluginViews = {
+  'midi-input': {
+    inputPanel: MidiInputPanel,
+  },
+}
```
funwithtriangles: Registering plugins in two places now, the engine and `pluginViews` which only `@hedron/desktop` cares about, the engine knowns nothing about this

packages/engine/src/HedronEngine/HedronEngine.ts
```diff
@@ -30,7 +30,7 @@ export class HedronEngine {
   }
 
   public registerPlugin(plugin: IPlugin) {
-    this.plugins.push(plugin)
+    this.plugins[plugin.id] = plugin
```
funwithtriangles, line 33: plugins are an object with plugin IDs (e.g. `midi-input`) rather than an array, just for easy lookup

packages/engine/src/plugins/Plugin.ts
```diff
@@ -1,10 +1,11 @@
...
-
 /**
  * Class type for a Plugin.
  */
 export interface IPlugin {
-  // new(engine: HedronEngine, useEngineStore: UseEngineStore): any;
+  /**
+   * Unique ID for plugin
+   */
+  id: string
```
funwithtriangles, line 8: Unique ID for plugin so it can be looked up wherever needed

packages/engine/src/plugins/Plugin.ts
```diff
@@ -17,9 +18,8 @@ export interface IPlugin {
...
 
   /**
-   * Gets the selected parameter view.
-   * @param param - The parameter with info.
-   * @returns The JSX element or undefined if this plugin does not have a visual element.
+   * Function to generate input options.
    */
-  getSelectedParamView?: (param: ParamWithInfo) => JSX.Element | undefined
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  generateInitialOptions: () => any
```
funwithtriangles, line 29: When an input is created, this function is called to give it whatever initial options it should have (e.g. default channel and note)

packages/midi-input/src/MidiInput.ts
```diff
@@ -0,0 +1,38 @@
...
+          event.channel === options.channel &&
+          event.note === options.note &&
+          event.value !== undefined
+        ) {
+          store.getState().updateNodeValue(input.targetNodeIds[0], event.value / 128)
+        }
+      })
+    })
+  }
+}
```
funwithtriangles: This is all the LFO input class is really doing! Just updating the store when a midi message comes in

packages/midi-input/src/MidiInputPanel.tsx
funwithtriangles, line 1: The view for the input. Needs a bit of tidying. Displays options for the input and has the MIDI learn functionality.

packages/midi-input/src/MidiInputPanel.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * A react component that displays the midi settings for a parameter
+ */
+export const MidiInputPanel = ({ input, engine }: IProps) => {
+  const [isLearning, setIsLearning] = useState(false)
+
+  // TODO: May not need this "as" if we have HedronEngineWithPlugin<MidiInput>
+  const plugin = engine.plugins['midi-input'] as MidiInput
+  const midi = plugin.midiManager
```
funwithtriangles: The component has access to whatever functionality it needs from the plugin by getting it from the engine. (e.g. `engine.plugins['my-plugin-id'].myCoolThing`)

packages/midi-input/src/MidiInputPanel.tsx
```diff
@@ -0,0 +1,118 @@
...
+                <EnumDropdown
+                  value={value}
+                  values={enumOptions[key]}
+                  onValueChange={(newVal) => {
+                    engine
+                      .getStore()
+                      .getState()
+                      .updateInputOptions(input.id, {
+                        [key]: newVal,
+                      })
```
funwithtriangles, line 79: I don't like this API. Probably just want it to be `engine.updateInputOptions`

packages/midi-input/src/MidiInputPanel.tsx
```diff
@@ -0,0 +1,118 @@
...
+    setIsLearning(true)
+    midi
+      .midiLearn()
+      .then((event) => {
+        if (!event) {
+          setIsLearning(false)
+          return
+        }
+
+        engine.getStore().getState().updateInputOptions(input.id, {
```
funwithtriangles, line 108: `runMidiLearn` now returns an `event` object rather than handling updates to the store internally

packages/desktop/src/renderer/engine.ts
cale-bradbury: Could we make it just the engine? Feel like that might be better for the injection of third party plugins, plugins that you want to work in both desktop and the web

packages/engine/src/plugins/Plugin.ts
cale-bradbury: hmm, what happens if a plugin wants to manage multiple input types? Maybe the midi one adds a velocity to `float` or a  pitch -> `Shot<number>`

packages/engine/src/plugins/Plugin.ts
cale-bradbury: I wonder if there is something good here with having a plugin override another, like someone makes a 'deluxe' lfo  plugin and still uses the id lfo. It can tap in and still work with a scene originally setup with regular lfo, and  deluxes changes at least somewhat fallback

packages/midi-test-app/src/App.tsx
```diff
@@ -29,11 +28,7 @@ function App() {
     $text('devices', getMidiDevices())
   }
 
-  const engine: HedronEngine = new HedronEngine()
-  const midi: Midi = new Midi(engine, () => {
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    return {} as any
-  })
+  const midi: Midi = new Midi()
```
cale-bradbury, line 31: nice

cale-bradbury (commented):
Just a few thoughts to help futureproof, can chat on thursday

funwithtriangles:
More points for meeting:
- have single `Midi` class instantiated inside `@hedron/desktop`, rename to `MidiManager`. Use in Clock too
- Breadcrumbs, cascading nodes. All options are nodes.
- Are params just nodes? do we need to have two type here?
- only one target node?
- inputs state flat?

packages/midi-manager/src/MidiManager.ts
funwithtriangles, line 1: @cale-bradbury I've reworked (and renamed!) this file, would be curious what you think. Basically going for the status byte as the value for the enum, which seems like a pretty stable option. It means that the `getMidiMessage` function is now very simple! It doesn't deal with unknowns but I feel like we can worry about that later on

funwithtriangles:
@cale-bradbury I'm ready to merge this as it's getting a bit monstrous. Since our meeting I just tidied up the `MidiManager` stuff a bit, particularly around the enum thingy. Let me know what you think! I still have [lots I want to do](https://github.com/nudibranchrecords/hedron/issues/505) around inputs, but it can wait for now!

PR #509, opened 2025-06-07
WIP Video Render

cale-bradbury:
From the new markdown...

# Experimental Features in Hedron

This document lists and describes experimental features available in Hedron. These features can be accessed via the developer console (f12).

### Saving an image
`saveFrame()` - Save a timestamped image.

### Saving an image sequence
`renderFrames(count: number, name: string, ffmpeg?: boolean)` - Save a series of images.
Images will be saved to the Documents folder, in a folder named name. It will overwrite any existing files. Each frame will be suffixed with its index. If `ffmpeg` is true, it will try to create a video using ffmpeg (must be installed and made accessible via the command line by the user).  
ffmpeg command as follows:
`ffmpeg -y -framerate 30 -i "${dirPath}/${name}-%d.png" -c:v libx264 -pix_fmt yuv420p -crf 18 "${videoPath}"`

### Testing a video loop
`resetEvery(seconds: number, jumpTo?: number)` - Loop after a given time
`cancelReset()` - Stop the looping
`jumpTo` is used to test the 'loop point' of a video, it will jump to the end, play through to the loop + the opposite of the jump time.
ie; `resetEvery(6, 5)` will play from seconds 5-6, then 6-1, then back to 5.
The way the video loops work is by passing large/negative `deltaTime` values to the sketches. `deltaTime` is a new top-level property that was added to the object passed on sketch `update`

### DIFF
--- a/EXPERIMENTAL_FEATURES.md
+++ b/EXPERIMENTAL_FEATURES.md
@@ -1,0 +1,21 @@
+# Experimental Features in Hedron
+
+This document lists and describes experimental features available in Hedron. These features can be accessed via the developer console (f12).
+
+## Rendering
+
+### Saving an image
+`saveFrame()` - Save a timestamped image.
+
+### Saving an image sequence
+`renderFrames(count: number, name: string, ffmpeg?: boolean)` - Save a series of images.
+Images will be saved to the Documents folder, in a folder named name. It will overwrite any existing files. Each frame will be suffixed with its index. If `ffmpeg` is true, it will try to create a video using ffmpeg (must be installed and made accessible via the command line by the user).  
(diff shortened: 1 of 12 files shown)


funwithtriangles:
This is so cool! The `renderFrames` wasn't working for me, because it was creating a new `debugScene` which was wiping everything in the scene. I have now added a `scene` property to the engine so can just reference that instead of making a new one. This is all temporary because I want to add proper scene functionality soon.

I'm happy to merge this! Maybe double check my changes are working for you?

cale-bradbury:
@funwithtriangles working good, feel free to merge it :)

PR #515, opened 2025-07-12
Sketch dispose function

cale-bradbury:
Adds an optional `dispose` function to sketches so they can do any cleanup they need.

Not beholden to the name if you think `destroy` would be better, just let me know, I'll update the md with whatever we go with

### DIFF
--- a/packages/engine/src/HedronEngine/importSketchModule.ts
+++ b/packages/engine/src/HedronEngine/importSketchModule.ts
@@ -23,8 +23,9 @@ export const importSketchModule = async (
     if ((await fetch(configPath)).status !== 200) {
       // No config file found
       // Try instancing the sketch, and call getConfig() on it
-      const tempModule = new module()
+      const tempModule = new module(undefined)
       config = tempModule.getConfig?.()
+      tempModule.dispose?.()
       if (!config) {
         return Promise.reject(
           `Sketch config not found: ${configPath} and no valid getConfig() function found in sketch`,
--- a/packages/engine/src/world/SketchManager.ts
+++ b/packages/engine/src/world/SketchManager.ts
@@ -16,13 +16,23 @@ type SketchInstance = {
   root?: Group
 
   getPasses?: (engineScene: EngineScene) => Pass[]
+
+  /**
+   * Called when the sketch is removed from the scene.
+   * It should clean up any resources, event listeners, or references it holds.
+   */
+  dispose(): () => void
 }
 
 export class SketchManager {
   private sketchInstances: { [id: string]: SketchInstance } = {}
 
-  private createSketch = (instanceId: string, module: SketchModule): SketchInstance => {
-    const sketch = new module(getDebugScene())
+  private createSketch = (
+    instanceId: string,
+    module: SketchModule,
+    scene: EngineScene,
+  ): SketchInstance => {
+    const sketch = new module(scene)
     if (sketch.root) {
       sketch.root.name = instanceId
     }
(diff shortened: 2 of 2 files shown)


packages/engine/src/world/SketchManager.ts
```diff
@@ -16,13 +16,23 @@ type SketchInstance = {
...
   private sketchInstances: { [id: string]: SketchInstance } = {}
 
-  private createSketch = (instanceId: string, module: SketchModule): SketchInstance => {
-    const sketch = new module(getDebugScene())
+  private createSketch = (
+    instanceId: string,
+    module: SketchModule,
+    scene: EngineScene,
+  ): SketchInstance => {
+    const sketch = new module(scene)
```
cale-bradbury, line 35: passing the EngineScene through instead, since we get it in the function prior anyways

packages/engine/src/HedronEngine/importSketchModule.ts
```diff
@@ -23,8 +23,9 @@ export const importSketchModule = async (
     if ((await fetch(configPath)).status !== 200) {
       // No config file found
       // Try instancing the sketch, and call getConfig() on it
-      const tempModule = new module()
+      const tempModule = new module(undefined)
```
cale-bradbury, line 26: Was running into gpu memory issues loading point clouds that weren't actually used since it was just calling getConfig(). if scene is undefined I only initialized what I needed to and skipped any loading

funwithtriangles: That's quite weird! Not sure how this makes any difference, isn't it still `undefined` when nothing is set? 😅 Fair enough though!

packages/engine/src/world/SketchManager.ts
funwithtriangles: Makes sense! We're gonna need a pattern like this later anyway when we have multiple scenes

packages/engine/src/world/SketchManager.ts
```diff
@@ -33,22 +43,25 @@ export class SketchManager {
...
+    const engineScene = getDebugScene()
+    const scene = engineScene.scene
     const oldSketch = scene.getObjectByName(instanceId)
 
     if (!oldSketch) {
       throw new Error(`couldn't find sketch to remove: ${instanceId}`)
     }
 
     scene.remove(oldSketch)
+    this.sketchInstances[instanceId]?.dispose?.(engineScene)
```
funwithtriangles, line 64: Nice to have `scene` handy to clear out some environment stuff after deleting a sketch that sets it up

```ts
  dispose({ scene }) {
    scene.background = null;
    scene.environment = null;
  }
  ```

funwithtriangles (approved):
This is great, thanks I just added one thing, passing the `engineScene` to `dispose`. Let me know if that makes sense for ya and I'll merge 👊

packages/engine/src/HedronEngine/importSketchModule.ts
cale-bradbury: Oh yeah it 100% is still undefined, I just wanted to write it explicitly for seeing what it is doing without needing to read the call signature. Was just talking about having some kind of marker for a real load vs config load in sketch

packages/engine/src/world/SketchManager.ts
cale-bradbury: True, sounds good, stops the sketch from needing to track internally.

Could see this causing issues with sketches that target the same thing, though that’s already the case. Maybe at some point we lock all the scene stuff down and have sketches flag themselves as a background modifier etc as to only apply the top-most. But I think that’s overkill for now, hack the planet.

funwithtriangles: Good point about the overwriting thing, although that would already be an issue if you have two sketches doing that. I agree with worrying about that later, maybe it won't really come up enough for us to care 👽

PR #516, opened 2025-07-12
Stop the computer from sleeping with Hedron open

cale-bradbury:
What the title says, tells the computer not to fall asleep when the app is open.

Stops any embarrassing interruptions to a show, since computers don't treat midi input as interactions like mouse/keyboard input.


funwithtriangles:
This is so cool, I didn't know this was possible and is clearly incredibly useful! Thanks!

I made one change: Only do the power save blocking when we're sending an output to another display. I figured the power saving feature is still important sometimes when working on a show and I leave Hedron running. Let me know if that makes sense and I'll merge this

cale-bradbury:
> I made one change: Only do the power save blocking when we're sending an output to another display.

Sounds good, send it!

PR #517, opened 2025-07-12
Fix error in midi-test-app

cale-bradbury:
Was failing to build because it was using the old MidiManager class


funwithtriangles:
Woo, thanks!

PR #524, opened 2025-07-18
Param groups

funwithtriangles:
Screenshot 2025-07-18 at 21 52 01

Adds way in config to group params into groups. I'm working with a sketch which is getting quite unwieldy with the params and wanted a way to keep it tidy on the screen (and in my head!) 

Still want to tidy up the design and have a few to-dos:

- [x] ~~toggle to show grouped params or in one big grid~~ overkill
- [x] collapsible groups (saved in state)
- [x] handle `undefined` group params
- [x] tidy up types (see TODO comments in code)
- [x] save/load state

Importantly is how we define this in the config. I went for the nested approach, because it feels the most natural. Once imported it gets flattened to keep things simple for developing.

@cale-bradbury how do you feel about this config format?

```ts
...
params: [
    {
      groupTitle: 'Mesh',
      groupKey: 'mesh',
      params: [
        {
          key: 'position',
          title: 'Position',
          valueType: 'vector3',
          defaultValue: [0, 0, 0],
        },
        {
          key: 'scale',
          title: 'Scale',
          defaultValue: 0.5,
          defaultMin: 0.00001,
          defaultMax: 4,
        },
      ],
    },
    {
      groupTitle: 'Material',
      groupKey: 'material',
      params: [
        {
          key: 'color',
          title: 'Color',
          valueType: 'rgb',
          defaultValue: [1, 1, 1],
        },
        {
          key: 'isWireframe',
          title: 'Wireframe',
          defaultValue: true,
          valueType: 'boolean',
        },
      ],
    },
    {
      key: 'someLooseParam',
      title: 'Param without a group',
      defaultValue: 0
    },
  ],
...
```


packages/engine/src/HedronEngine/importSketchModule.ts
```diff
@@ -1,7 +1,39 @@
...
+const processConfig = (config: SketchConfigRaw): SketchConfig => {
+  const groupInfo: SketchConfig['groupInfo'] = []
+
+  let groupIndex = -1
+  const flattenedParams = config.params.flatMap((param) => {
+    if ('params' in param) {
+      groupIndex++
+
+      groupInfo[groupIndex] = {
+        groupTitle: param.groupTitle ?? `Group ${groupIndex}`,
```
cale-bradbury:
```suggestion
        groupTitle: param.groupTitle ?? `${config.title} ${groupIndex}`,
```

cale-bradbury (approved):
Looking good! Great addition, always frustrating when my params aren't in groups of 3

packages/engine/src/HedronEngine/importSketchModule.ts
funwithtriangles: What's the idea here? Doesn't make much sense to me to have the sketch name as part of the group name 🤔

packages/desktop/src/shared/types.ts
```diff
@@ -5,5 +5,11 @@ export interface ProjectData {
   engine: EngineData
   app: {
     sketchesDir: string
+    // TODO: activeSketchId should be part of the save state
+    // but causing errors when Hedron opens directly on a sketch
+    // activeSketchId: string | null
+    selectedNodes: { [sketchId: string]: string }
+    selectedInputs: { [inputId: string]: string }
+    openedParamGroups: { [sketchId: string]: Record<number, boolean> }
```
funwithtriangles, line 13: These bits of state are now saved to the project file, so when you load your project back up, it has you where you left off in terms of opened/closed stuff

packages/engine/src/HedronEngine/importSketchModule.ts
```diff
@@ -1,7 +1,69 @@
...
+  }
+
+  const processedConfig: SketchConfigImported = {
+    ...config,
+    title: config.title ?? fallBackTitle,
+    params: flattenedParams,
+    groupInfo,
+  }
+
+  return processedConfig
```
funwithtriangles, line 64: this is adding some type safety to how we import stuff. We have an expected raw config format and an expected imported format.

packages/engine/src/HedronEngine/importSketchModule.ts
```diff
@@ -17,35 +79,31 @@ export const importSketchModule = async (
...
-      config.title = moduleId
-    }
-
-    for (const param of config.params) {
-      if (!param.valueType) {
-        // @ts-expect-error - valueType is NOT optional in SketchConfigParam, but is optional for users in config.ts
-        param.valueType = NodeTypes.Number
-      }
+      config = processConfig(configModule.default as SketchConfigRaw, processConfigOptions)
     }
```
funwithtriangles, line 107: all of this is handled by `processConfig` and is type safe now

packages/engine/src/store/types.ts
```diff
@@ -123,15 +124,43 @@ export type SketchConfigParam =
...
+  params?: (SketchConfigParam | SketchConfigGroup)[]
+}
+
+export interface SketchConfigImported {
+  title: string
+  description?: string
+  params: SketchConfigParamImported[]
+  groupInfo: { groupTitle: string }[]
+}
+
```
funwithtriangles, line 159: We now have types for both the raw config files that users write, and the imported version that Hedron deals with after importing the file.

funwithtriangles:
@cale-bradbury this is all done now, I'll merge it soonish, but feel free to check it out before then. I did a few things cleaning up types and also how the project is saved so a few more little state things get saved. You can see how this works by opening up `project.json` in the example project. The `solid` sketch uses param groups.

PR #529, opened 2025-07-24
Webgpu postprocessing WIP

funwithtriangles:
First attempt at getting WebGPU post processing working with sketches. It seems to be working but I'm not entirely convinced by it all just yet!

One big concern is that this doesn't care about sketch order at all and I feel like even if we have some sort of defined sketch order, the passes could still be added in any random order. For instance `reimportSketchModuleAndReloadSketches` would add passes in the wrong order, etc. 

~We probably want some single method that clears passes and goes through all sketches to re-add passes. We'd do this on init but also whenever any file change happens.~

~Maybe the neatest option is to set a flag like `passesNeedReset` and we just do it once before rendering and then turn off the flag. Then any sort of thing can trigger multiple times but we only do the pass stuff once.~

EDIT: Done!

todo:

- [x] build failing
- [x] check webGL/webGPU both functioning
- [x] example project should be webGL
- [x] eslint/ts config check over


.eslintrc.cjs
```diff
@@ -30,5 +30,11 @@ module.exports = {
     react: {
       version: 'detect',
     },
+    'import/resolver': {
+      typescript: {
+        alwaysTryTypes: true,
+        project: './tsconfig.json',
+      },
+    },
```
funwithtriangles: Was having issues with types and eslint complaining. Not sure why this wasn't needed before and suddenly now. Want to investigate further before merging, I hate having loads of random config I don't understand, especially if it's not needed!

tsconfig.json
```diff
@@ -2,6 +2,9 @@
   "extends": "@tsconfig/recommended/tsconfig.json",
   "exclude": ["node_modules"],
   "compilerOptions": {
-    "noImplicitAny": true
+    "noImplicitAny": true,
+    "module": "esnext",
+    "moduleResolution": "bundler",
+    "allowSyntheticDefaultImports": true
```
funwithtriangles: same as comment above, config i don't really understand

.eslintrc.cjs
funwithtriangles: I have reverted this change for now, might still be useful to try in a different PR  but not really the point of this PR

tsconfig.json
funwithtriangles: I have reverted this change for now, might still be useful to try in a different PR  but not really the point of this PR

funwithtriangles:
Merging this for now as it works in a simple way if you have a single sketch handling your postprocessing stuff!

PR #531, opened 2025-07-25
LFO Input with big refactor

funwithtriangles:
Summary of what this PR does:
- Moves extra store-related hooks and container components to `ui-core`. This is so that plugin packages can make use of store-related stuff without any issue. In order for this to all work properly, we're now using  modified `useEngineStore` and `useAppStore` hooks that use context so that the store is always accessible
- Moves the app store to its own package so that `ui-core` can import the correct types. Felt very wrong importing `@hedron/desktop` into `ui-core` and I was getting cyclical dependency warnings
- Input options are now nodes in the store rather than properties on the input
- Nodes now contain all of their config (e.g. `title`, `options`). Originally stuff like options array for enums, and titles of params were kept only in the sketch module part of the store. This felt like the right way to do things at the time, because less duplication (e.g. we don't have the same array of enum options all over the project file). However it's now clear that this is going to get trickier and trickier to manage over time, as we have different node types but want to reuse the same UI components to display them. 
- Input plugins have a slightly new structure, the config is defined in the same way sketches params are
- LFO input
- MIDI Input updated to follow new structure

TODO:

- [x] Use nodes for options
- [x] type fixes 
- [x] rework MIDI input
- [x] display node based options, create reusable `addNode` and components for this
- [x] shared hooks and engine store... https://chatgpt.com/share/68aeb28d-0794-8012-b6bd-656a80417578
- [x] move all node config to node instances
- [x] fix `sliderMin` and `sliderMax` regression
- [x] fix sketch updates after node config changes (e.g. pass options/title/groupIndex to nodes)
- [x] dropdowns for some options
- [x] cleanup vite config
- [ ] ~enable/disable options as generic option?~ https://github.com/nudibranchrecords/hedron/issues/553
- [ ] ~delete inputs (for MIDI also, could be same bottom toolbar as enable/disable)~ https://github.com/nudibranchrecords/hedron/issues/553
 - [x] skip missing nodes (when a sketch/param is deleted and didn't clean up properly)
 - [x] min/max midi range fixes (see commit in `awamu-2025`)
 - [x] package readme

~~Cherry Pick things (little bugs I'm finding on my messy  `awamu-2025` branch that i'll need to re-implement later). Probs best in another PR!~~ Moved to #547 
- [x] `ActiveSketch` `SketchParams` component needs unique key
- [x] dodgy scroll fix for params
- [x] update electron builder (fixes issue with mac)
- [x] tidy up tick loop
- [x] break up switch case for loop
- [ ] ~show hide certain option nodes in the LFOInputPanel depending on node value type (e.g. boolean can have a `pulseWidth` option, special enum blacklist)~ https://github.com/nudibranchrecords/hedron/issues/563
- [x] check if need this on MidiInput  `if (input.type !== 'midi') return`
- [ ] ~special context/hook for options~ https://github.com/nudibranchrecords/hedron/issues/564


packages/desktop/src/renderer/components/App/App.tsx
```diff
@@ -1,32 +1,39 @@
...
   return (
-    <div className={c.wrapper}>
-      <div className={c.left}>
-        <Viewer />
-        <div className={c.widgetStrip}>
-          <PerformanceStats />
-          <div className={c.widgetItem}>
-            <GlobalClock />
+    <AppStoreProvider value={appStore}>
+      <EngineStoreProvider value={engineStore}>
```
funwithtriangles, line 16: We now have Providers that provide the two stores via React context. This means we can use hooks such as `useEngineStore` without needing a reference to the global store instance. It means that all components that consume the store (including those from `@hedron/ui-core`) now work with plugins.

packages/desktop/src/renderer/components/hooks/useActiveSketchParams.ts
```diff
@@ -27,17 +27,9 @@ export const useActiveSketchParams = () => {
...
 
-      if (!paramConfig) {
-        console.warn(`No param config found for node ${id} in sketch ${activeSketch.id}`)
-        return
-      }
-
-      const title = paramConfig?.title ?? paramConfig?.key
-
-      const isUngrouped = paramConfig.groupIndex === null
-      const groupIndex = isUngrouped ? module.config.groupInfo.length : paramConfig.groupIndex!
```
funwithtriangles, line 40: We're no longer looking up the title of a param via the global config. All properties on params belong to each param in the store. This definitely means the project file will be bigger (repeated information) but it also simplifies the code a lot, making plugin development easier. 

Funnily enough it won't actually make much of a difference to the project file sizes because we were accidentally including these properties anyway!

packages/engine/src/HedronEngine/HedronEngine.ts
```diff
@@ -50,6 +52,10 @@ export class HedronEngine {
 
     this.onFrameStart = params?.onFrameStart
     this.onFrameEnd = params?.onFrameEnd
+
+    if (params.clock) {
+      this.clock = params.clock
+    }
```
funwithtriangles, line 58: engine can optionally have a `clock` property now, used by the LFO input (and whatever else may want to use it)

packages/desktop/src/renderer/engine.ts
```diff
@@ -13,17 +13,17 @@ export const engine = new HedronEngine({
...
   onFrameEnd: performanceMonitor.end,
   rendererType: import.meta.env.HEDRON_RENDERER_TYPE ?? 'webgl',
+  clock,
 })
 
 export const engineStore = engine.getStore()
 
-export const useEngineStore = <T>(selector?: (state: EngineStateWithActions) => T) => {
-  return useStore(engineStore, selector!)
-}
```
funwithtriangles, line 22: `useEngineStore` is no longer coupled to a specific store and has been moved to `@hedron/ui-core`. The new version of this hook uses React context and needs to have `<EngineStoreProvider>` as an ancestor component (e.g. at the root of the React tree), with the store handed there.

packages/engine/src/plugins/Plugin.ts
```diff
@@ -23,8 +32,85 @@ export interface IPlugin {
...
+    ? // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      K['options'] extends readonly any[]
+      ? K['options'][number]['value']
+      : K['defaultValue']
+    : K['valueType'] extends 'number'
+      ? number
+      : K['valueType'] extends 'boolean'
+        ? boolean
+        : K['defaultValue']
+}
```
funwithtriangles: This mad utility type is a way of parsing the options node config so we have nice types when using the `optionNodes` object. This is a Github Copilot generated thing, will be revisiting this soon when typing sketch files and their params.

packages/engine/src/plugins/Plugin.ts
```diff
@@ -23,8 +32,85 @@ export interface IPlugin {
...
+
+/**
+ * Plugin helper that loops through all inputs and handles those of a specific type.
+ * Gets and hands back option nodes, the target node for the input, and its value.
+ * @param storeState The engine store state.
+ * @param inputType The type of input to handle (e.g. "midi")
+ * @param callback The callback to execute for each input of the specified type.
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const handleEachInput = <T extends readonly any[]>(
```
funwithtriangles, line 83: This helper function can be used by every input plugin

packages/engine/src/store/shared/addNode.ts
funwithtriangles, line 1: the `addNode` store function has been generalised, no longer needing a `sketchId`

packages/engine/src/store/types.ts
```diff
@@ -19,52 +19,52 @@ export interface NodeBase {
...
 }
 
-export enum NodeTypes {
-  Number = 'number',
-  Boolean = 'boolean',
-  String = 'string',
-  Enum = 'enum',
-  Vector3 = 'vector3',
-  RGB = 'rgb',
-}
```
funwithtriangles, line 29: This TS enum has been removed, as it was causing some TS issues and we don't actually need it. We still have complete type safety because every type of node has the `valueType` associated with it.

This is also a necessary change for later, when we want users to be able to type their config files. With this enum, they'd need to import `NodeTypes` just to give each param in their config a `valueType`.

packages/engine/src/store/types.ts
```diff
@@ -80,9 +80,12 @@ export type Nodes = { [key: string]: Node }
 
 export type NodeValue = number | boolean | string
 export type NodeValues = { [key: string]: NodeValue }
+export type NodeValueType = Node['valueType']
```
funwithtriangles, line 83: We still have `NodeValueType`, which evaluates to a union of all node value types 

```ts
 "string" | "number" | "boolean" | "enum" | "vector3" | "rgb"
```

packages/lfo-input/src/LFOInput.ts
```diff
@@ -0,0 +1,188 @@
...
+const TAU = Math.PI * 2
+const lerp = (v0: number, v1: number, t: number) => (1 - t) * v0 + t * v1
+
+export class LFOInput implements IPlugin {
+  public readonly id = 'lfo-input'
+  public readonly name = 'LFO Input'
+  public readonly inputType = 'lfo'
+  public readonly description =
+    'Generates LFO waves (e.g. sin, square, sawtooth) as inputs for params.'
+  public readonly optionNodesConfig = [
```
funwithtriangles, line 31: Input plugins might want option nodes. This config is the same format as the sketch params config. These nodes will be generated for any new input created

packages/lfo-input/src/LFOInput.ts
```diff
@@ -0,0 +1,188 @@
...
+
+    const store = engine.getStore()
+
+    const tick = () => {
+      requestAnimationFrame(() => {
+        const storeState = store.getState()
+        handleEachInput<typeof this.optionNodesConfig>(
+          storeState,
+          'lfo',
+          ({ input, optionNodes: opts, targetNode }) => {
```
funwithtriangles: We have access to `optionNodes` here, using the new helper function, `handleEachInput`

packages/lfo-input/src/LFOInputPanel.tsx
funwithtriangles, line 1: The component part of the plugin can be very simple now!

packages/midi-input/src/MidiInput.ts
```diff
@@ -1,18 +1,21 @@
...
-  note: number
-  type: MidiMessageType
+const noteLetters = ['C', 'C#', 'D', 'D#', 'E', 'F', 'F#', 'G', 'G#', 'A', 'A#', 'B']
+
+const midiNotes: string[] = new Array(128)
+
+for (let i = 0; i < 128; i++) {
+  const letter = noteLetters[i % noteLetters.length]
+  const octave = Math.floor(i / noteLetters.length) - 1
+  midiNotes[i] = `${i} - ${letter} (${octave})`
```
funwithtriangles, line 23: @cale-bradbury it would be great if you could test the MIDI options and let me know if you're okay with the note names like this? It's the same as the old hedron

packages/ui-core/src/containers/Param/Param.tsx
funwithtriangles, line 1: `Param` is now a generalised container component that uses the engine store based on context

packages/ui-core/src/hooks/storeHooks.ts
funwithtriangles, line 1: these new store hooks are using React context instead of being coupled directly with a single store instance. Gives us more flexibility and allows third parties and other packages in this monorepo to use these hooks, knowing that the main desktop app uses a provider

packages/app-store/README.md
```diff
@@ -0,0 +1,15 @@
...
+- Node and input selections
+- Parameter group states
+- Sketches directory path
+- Global dialog state
+- Save file management
+- Project persistence
+
+## Usage
+
+Most usage would actually be via `useAppStore` imported from `@hedron/ui-core`
```
cale-bradbury, line 15: Should `useAppStore` come from the app-store package instead?

packages/app-store/src/appStore.ts
```diff
@@ -0,0 +1,146 @@
...
+import { EngineData } from '@hedron/engine'
+
+export type DialogId = 'sketchModules'
+
+export interface SaveItem {
+  title: string
+  path: string
+  date: number
+  numScenes: number
+  numSketches: number
```
cale-bradbury, line 15: do we need to know the count as it's own prop? Wouldn't the save item contain the scene/sketch data instead, and you could just `.length` the arrays

packages/desktop/src/renderer/components/App/App.tsx
```diff
@@ -1,32 +1,39 @@
...
+            <Viewer />
+            <div className={c.widgetStrip}>
+              <PerformanceStats />
+              <div className={c.widgetItem}>
+                <GlobalClock />
+              </div>
+              <div className={c.widgetItem}>
+                <VideoControls />
+              </div>
+              {/* <Audio /> */}
```
cale-bradbury, line 28: Will probably want these coming dynamically from the plugins in the future

packages/desktop/src/renderer/engine.ts
```diff
@@ -13,17 +13,17 @@ export const engine = new HedronEngine({
...
 engine.registerPlugin(new MidiInput(engine))
+engine.registerPlugin(new LFOInput(engine))
```
cale-bradbury, line 22: any thoughts on how to get these auto-picked up, wouldn't block it for this, I'm fond of decorators myself.

Also, wouldn't something like the lfo (and maybe midi in some curcumstances) be needed when running in a webpage, and should be handled in the engine instead of the desktop app (or maybe a third thing, plugin manager...)

packages/engine/src/HedronEngine/HedronEngine.ts
cale-bradbury: One thing I noticed regarding the clock, is how update still gets called when it isn't running, wondering if that shouldn't be the case, the the app fully pauses when the clock isn't running. Thinking it would be good to have one source of truth regarding time that drives everything

packages/engine/src/store/types.ts
```diff
@@ -177,17 +182,18 @@ export interface SketchModuleItem {
...
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export type InputOptions = any
+export type InputOptionNodesConfig = readonly SketchConfigParam[]
 
-export interface Input<Options = InputOptions> {
+export interface Input {
   id: string
   title: string
   type: 'midi' | 'gamepad' | string
```
cale-bradbury, line 192: Feel like maybe this should just be string? since there could end up being all sorts of types coming via plugins?

cale-bradbury: I guess type being the way for a plugin to know the input belongs to it (or perhaps that it wants to modify it in the case of say a plugin that mocks midi messages, etc?)

cale-bradbury (commented):
Have yet to fully review, just got to the end of engine. Will try to finish later this weekend

packages/engine/src/HedronEngine/HedronEngine.ts
funwithtriangles: this does sound like a nice feature but I can also understand why some people might want to have certain little animations running if they pause (thinking web projects) Would be great to have this as some setting somewhere! Also having a special delta as part of the `update` parameters that is tied to the clock speed would be pretty nice too

packages/app-store/README.md
funwithtriangles: I can see the logic in wanting to keep these together. My main reasoning for keeping `useAppStore` and `useEngineStore` in the `ui-core` is that they are only being used for UI, and have React as a dependency.

packages/app-store/src/appStore.ts
funwithtriangles: This type is for saved items which are stored in `localStorage`. When Hedron first loads, this is how we're displaying a list of previous projects on the home screen. It doesn't load the whole project for each, just this little object of meta data

packages/app-store/README.md
cale-bradbury: Okay, we will see if they are wanted in other plugins

packages/engine/src/HedronEngine/HedronEngine.ts
cale-bradbury: Maybe like per-sketch meta-settings (`updateWhenPaused`, `active`, etc.)

packages/lfo-input/src/LFOInput.ts
cale-bradbury: Love this pattern, would be cool if it's also the case for top-level settings of any node, like those for audio smoothing/falloff, etc

packages/lfo-input/src/LFOInput.ts
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    const store = engine.getStore()
+
+    const tick = () => {
+      requestAnimationFrame(() => {
+        const storeState = store.getState()
+        handleEachInput<typeof this.optionNodesConfig>(
+          storeState,
+          'lfo',
```
cale-bradbury:
```suggestion
          this.inputType,
```

packages/lfo-input/src/LFOInput.ts
cale-bradbury: Would be nice to move this into a class level function, both to help with tab depth, and for the comment

packages/lfo-input/src/LFOInput.ts
```diff
@@ -0,0 +1,188 @@
...
+          ({ input, optionNodes: opts, targetNode }) => {
+            // TODO: This can be handled by `onInput` once we have `isEnabled` as a generic option
+            if (!opts.isEnabled) return
+
+            const delta = clock.beatDelta * opts.frequency * TAU + opts.phase
```
cale-bradbury:
```suggestion
            const delta = (clock.beatDelta * opts.frequency + opts.phase) * TAU
```
have phase complete a full loop

packages/lfo-input/src/LFOInput.ts
```diff
@@ -0,0 +1,188 @@
...
+
+            let value: NodeValue | null = null
+
+            switch (targetNode.valueType) {
+              case 'enum':
```
cale-bradbury: I think it would be good to move all of the switch cases into top-level functions, and to have enum and boolean take  min/max into account (entries or percent of the list, and a pulse-width size, respectively)

packages/lfo-input/src/LFOInput.ts
```diff
@@ -0,0 +1,188 @@
...
+  HedronEngine,
+  InputOptionNodesConfig,
+  IPlugin,
+  NodeValue,
+} from '@hedron/engine'
+
+const TAU = Math.PI * 2
+const lerp = (v0: number, v1: number, t: number) => (1 - t) * v0 + t * v1
+
+export class LFOInput implements IPlugin {
```
cale-bradbury, line 25: This class looks pretty simple, It looks like a great example, doesn't feel like there is extraneous boilerplate. Feeling good about this system

packages/lfo-input/src/LFOInputPanel.tsx
```diff
@@ -0,0 +1,29 @@
...
+  input: Input
+  // TODO: This can be typed as something like HedronEngineWithPlugin<MidiInput>
+  engine: HedronEngine
+}
+
+const Item = ({ nodeId }: { nodeId: string }) => {
+  const param = useEngineStore((state) => state.nodes[nodeId])
+
+  return <Param param={param} />
+}
```
cale-bradbury: Would be nice if this could just come from ui-core `<NodeParam..>` perhaps

packages/lfo-input/src/LFOInputPanel.tsx
```diff
@@ -0,0 +1,29 @@
...
+ */
+export const LFOInputPanel = ({ input }: IProps) => {
+  return (
+    <div>
+      <ControlGrid className="mb-xl">
+        {input.optionNodeIds.map((id) => (
+          <Item key={id} nodeId={id} />
+        ))}
+      </ControlGrid>
+    </div>
```
cale-bradbury: This could also be a good one to come from ui-core for getting your don't-care-default-display.
`<OptionGrid>` ?

packages/lfo-input/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
```
cale-bradbury, line 1: Would be good to have a root package script to generate a new sub-package to add all the config files, etc. Maybe could even add it to the list desktop consumes.

cale-bradbury (commented):
Got through LFO 😓 It's looking like an elegant plugin system so far

packages/midi-input/src/MidiInput.ts
cale-bradbury: Yup, good with the names like this. Might be cool to have a little button to pop out a keyboard or some gui in the future instead of the default enum dropdown, but this is great for now.

packages/midi-input/src/MidiInput.ts
```diff
@@ -21,80 +24,103 @@ export class MidiInput implements IPlugin {
...
-          event.type === options.type &&
-          event.value !== undefined
-        ) {
-          const node = storeState.nodes[input.targetNodeId]
-          const nodeVal = storeState.nodeValues[input.targetNodeId]
+      handleEachInput<typeof this.optionNodesConfig>(
+        storeState,
+        'midi',
+        ({ input, optionNodes: opts, targetNode, targetNodeValue }) => {
+          if (input.type !== 'midi') return
```
cale-bradbury: can ditch this line now because it's filtered by handleEachInput, correct?

packages/midi-input/src/MidiInput.ts
```diff
@@ -21,80 +24,103 @@ export class MidiInput implements IPlugin {
...
+                      targetNode.options[
+                        Math.floor((event.value / 127) * (targetNode.options.length - 1))
+                      ].value
+                    break
+                  }
+                }
+                break
+              }
+              case 'number': {
+                value = (event.value / 127) * (sliderMax - sliderMin) + sliderMin
```
cale-bradbury: I don't think it is appropriate to use slider min/max here. That is a top level param modifier, and I feel like inputs should be expected to just output a normalized value (which could potentially have it's own min/max like with lfo)

packages/midi-input/src/MidiInputPanel.tsx
```diff
@@ -1,87 +1,85 @@
...
+              state.updateNodeValue(nodeId, event.channel)
+              break
+            case 'note':
+              state.updateNodeValue(nodeId, event.note)
+              break
+            case 'type':
+              state.updateNodeValue(nodeId, event.type)
+              break
+          }
+        })
```
cale-bradbury: this feels a little clunky, would be nice if it was a one liner for each key you want to modify, not really sure what that would look like. `state.updateInputOptionNodeValue(input, 'channel', event.channel)`, or maybe `input.updateOptionValue('channel', event.channel)`

could even have a bulk one that takes an object and updates any keys that have a match, so you could theoretically `input.bulkUpdateOption(event)`

packages/ui-core/src/hooks/storeHooks.ts
cale-bradbury: I'm curious if providers would be how you could signal where Hedron is running. Like if it's in web the hedron bootstrap could have a WebStoreProvider instead of an AppStoreProvider (or thats the type of AppStoreProvider that is provided)? Maybe the app one could provide (or just signal it is safe to use) `fs` etc.? I don't really know how the provider architecture works.

cale-bradbury (commented):
Okay looking good, just a few notes and confusions, if you want to chat on thursday again for clearing things up, and faster flow, hmu

packages/engine/src/HedronEngine/HedronEngine.ts
funwithtriangles: ah yeah nice!

packages/lfo-input/src/LFOInput.ts
funwithtriangles: Definitely should be the same pattern for the audio stuff!

packages/ui-core/src/hooks/storeHooks.ts
funwithtriangles: No need for providers with the web version, it is completely React free 😎 

The providers are just handing the store down to UI components so that hooks can make use of them for controls, etc.

packages/desktop/src/renderer/engine.ts
funwithtriangles: If you take a look at my web example PR (https://github.com/nudibranchrecords/hedron/pull/559), you still have to register the plugins you want. This is to keep things slim for web projects, only bringing in the parts you need.

For the desktop version we'll eventually want some extra layer where custom plugins can be automatically pulled in from the file system, etc.

packages/ui-core/src/hooks/storeHooks.ts
cale-bradbury: So hypothetically, say I wanted to make a website that would use some amount of hedron UI (or even a web based version of the desktop app), or say a phone version you could use to control hedron from in the crowd, casting to the projector. None of this needs to be a priority atm, just thinking to the future

cale-bradbury: (to be fair, for a mobile version, would likely just use an osc app with hedron running on a computer tucked away somewhere, but in theory would be cool if you always had viz ready to go in your pocket)

packages/engine/src/store/types.ts
funwithtriangles: yeah this type is kind of useless anyway once you add `string` to the union 😛

packages/lfo-input/src/LFOInput.ts
funwithtriangles: Not sure I fully understand, do you mean like a method on the plugin? What do you mean with tab depth? 😅

packages/lfo-input/src/LFOInput.ts
funwithtriangles: whoops, nice catch! Thanks!

packages/lfo-input/src/LFOInput.ts
funwithtriangles: Oh yeah that's a nice idea. Although I think we can do a bit better than that and have specific options for these value types, with proper names (e.g. "pulse width"). Those nodes can be shown/hidden depending on the valueType of the target node (which we should be able to do quite easily over in `LFOInputPanel`).

I'd say that's something for a later date though!

packages/ui-core/src/hooks/storeHooks.ts
funwithtriangles: Ah I gotcha! Technically you should be able to build out a simple mini interface with the packages we already have.

packages/lfo-input/src/LFOInputPanel.tsx
funwithtriangles: Great suggestion. I've gone one step further: `Param` now just takes a `paramId` so it can be used directly below.

packages/lfo-input/src/LFOInputPanel.tsx
funwithtriangles: We could have this, and I can see the temptation, but there's always a line that needs to be drawn. Things start getting a bit annoying to maintain when you start wrapping up simple component patterns like this. At some point you start getting edge cases that require props to be passed down to adjust (e.g. some margin adjustment based on some parent panel's spacing)  and in the end its better to keep things open like this.

packages/lfo-input/tsconfig.json
funwithtriangles: Agree that it's still a bit daunting. Although I'm a bit nervous to start thinking about scripts like this because they can get outdated quickly. Isn't another option just to copy the package most similar to the one you want to create?

packages/midi-input/src/MidiInputPanel.tsx
funwithtriangles: Yeah could also be a hook. Something like:

```ts

// Either we manually pass this hook `input.optionNodeIds` or it gets it from context somehow
const updateOptionByKey = useUpdateOptionNodeByKey()
 
...

updateOptionByKey('channel', event.channel)
```

packages/midi-input/src/MidiInput.ts
funwithtriangles: This is a really tricky one and i've gone back and forth on it a few times. My feelings are:

- LFOs have their own min/max because its not a human controlling it and you may want many settings of LFO for the same param
- MIDI inputs follow the `sliderMin` and `sliderMax` because if you've set them for a virtual slider already, you _probably_ want the same thing once you've mapped it over to the midi device. Could definitely give MIDI inputs their own min/max though (and I'd suggest we default them to whatever `sliderMin` and `sliderMax` is when the input was created)

packages/lfo-input/src/LFOInput.ts
funwithtriangles: This is now done, let me know what you think of the solution...

PR #561, opened 2025-09-13
LFO Hotfix (use sliderMix/Max values)

cale-bradbury:
Add node uses sliderMin/sliderMax when adding to the state


funwithtriangles:
Good catch, thanks!

PR #566, opened 2025-09-14
Stars sketch update

cale-bradbury:
Updated the stars sketch to provide more control, you can set particle count, colors, range. Uses deltaTime now as well for consistency across FPS.


funwithtriangles:
Hey, this is cool! Thanks for taking this sketch to the next level. I reverted one commit you made regarding the engine, cos it was breaking the types

PR #568, opened 2025-09-20
Intro Improvements

cale-bradbury:
Before: Projects would get cut off, buttons off-screen
hedron-before

After: Very demure, very mindful
image

The full paths are now clickable, opening the folder, adding a nice function to do so
Saved project data now goes through a sanitization step to remove the path in the title, and theoretically checks that the paths are valid (the bottom 2 projects of my top picture were only correct titles because I had previously tried the changes)


packages/app-store/src/appStore.ts
```diff
@@ -60,6 +60,32 @@ export type CustomSetState = (
...
+      const filename = item.title.split(/[\/\\]/).pop() || item.title
+      // Remove extension if present
+      const titleWithoutExt = filename.replace(/\.[^/.]+$/, '')
+      return { ...item, title: titleWithoutExt }
+    }
+    return item
+  })
+  // remove paths that do not exist anymore
+  // this doesn't actually work yet, because I couldn't figure out how to get fs on the global object before the store is created
+  const fs = window['__HEDRON'].fs
```
cale-bradbury: I tried getting it to add fs to the __HEDRON object in the desktop package, but I think this was always running first? Regardless, even trying to add it after the window was created, the object never displayed fs on it, despite using the same code as in globalVars (blend existing object, don't overwrite)

cale-bradbury: Maybe this is the wrong way to go about it, and I should fix the values elsewhere, but I'd still like to get fs available as `typeof fs | undefined` so that sketches can check for it and do stuff if so (like saving out data files, etc)

packages/app-store/src/appStore.ts
```diff
@@ -60,6 +60,32 @@ export type CustomSetState = (
...
+    // Check if the title looks like a path
+    if (item.title && (item.title.includes('/') || item.title.includes('\\'))) {
+      // Extract filename without extension from the path
+      const filename = item.title.split(/[\/\\]/).pop() || item.title
+      // Remove extension if present
+      const titleWithoutExt = filename.replace(/\.[^/.]+$/, '')
+      return { ...item, title: titleWithoutExt }
+    }
+    return item
+  })
```
funwithtriangles: So I think you're trying to solve this problem in the wrong place. 

This cleanup shouldn't be necessary as we should always be extracting this filename whenever the user saves a file (see `packages/desktop/src/renderer/handlers/fileHandlers.ts` line 87). I don't have that problem on my mac, however clearly something is broken on Windows...

We're using `path.basename(response.savePath)` to grab that filename, and as an improvement we can even do `path.basename(response.savePath, '.json')` to remove the `.json` extension. This _should_ be OS specific, and work across both Windows and POSIX systems. Not exactly sure what's going on here??

packages/app-store/src/appStore.ts
funwithtriangles: This will never work as fs is not available to the Electron renderer. I think we need to build a little middle layer to allow users to save files.

packages/desktop/src/main/handlers/openFolder.ts
```diff
@@ -0,0 +1,37 @@
...
+    await shell.openPath(normalizedPath)
+    return { success: true }
+  } catch (err) {
+    console.error('Error opening folder:', err)
+    return {
+      success: false,
+      error: err instanceof Error ? err.message : 'Unknown error opening folder',
+    }
+  }
+}
```
funwithtriangles, line 41: This is a great feature, awesome stuff!

packages/engine/src/hedronTypes.ts
```diff
@@ -0,0 +1,23 @@
...
+    POSTPROCESSING: typeof POSTPROCESSING
+  }
+  fs?: any // fs is only available when running on desktop
+}
+
+declare global {
+  interface Window {
+    __HEDRON: HedronGlobal
+  }
+}
```
funwithtriangles: I'm not so keen on promoting the hacky `__HEDRON` namespace. The idea behind it is just as a middle step to mock package imports in sketches (e.g. `import { MeshNormalMaterial } from 'three'`). The moment we start typing this and tacking on extra properties I feel like we're heading down a messy path...

funwithtriangles (changes requested):
Hey Cale, thanks for these improvements, this is really cool and there are some parts of this I'd love to merge right away:
- Clicking on the folder names to open file explorer 😻 
- CSS fixes 🎨 

However there are some other parts of this PR that don't quite make sense to me:
- Exposing `fs` and extending the hacky global namespace 😬 
- `cleanupSaveListTitles` seems like the wrong solution to the problem 🤔 Not sure what's broken on your machine that you're having this problem in the first place...

I have left comments to try my best to explain myself.

funwithtriangles:
For the pathname issue, it almost certainly is because we're not actually using Node's `path`, we're using some browser thingy: https://www.npmjs.com/package/path-browserify

Ideally we don't use `path-browserify` and find something that does a better job with Windows/POSIX

Edit: I think I already [fixed it](https://github.com/nudibranchrecords/hedron/pull/574)

PR #569, opened 2025-09-20
Audio Input Plugin

cale-bradbury:
Still WIP, but can take feedback on what's there

- Moves all the existing audio logic into a plugin
- Bands now have bell curves applied
- Visualizer to set band freq and Q-factor
- Change audio devices
image


packages/engine/src/HedronEngine/HedronEngine.ts
```diff
@@ -160,6 +210,17 @@ export class HedronEngine {
     return stripForSave(this.store.getState())
   }
 
+  /**
+   * Recreates global option nodes for all registered plugins
+   * This should be called after loading a project or resetting the store
+   */
+  public recreateGlobalOptionNodes() {
```
cale-bradbury: on project load, the state gets cleared so recreating the global options. Not sure if we want these to go beyond the project, maybe we just wait to register them until load, but then we need to hide the widget strip until loaded

cale-bradbury: Maybe it would actually be good to hide all widgets on load, we could use the canvas for a demoscene, and make use of all the space to get you into your work faster, fit more projects, select sketch folder, news/updates

funwithtriangles: I think I'm a bit confused on what these two methods are doing and why we need it compared to how its working for the LFO input right now, probably worth having a chat soon to discuss it all?

funwithtriangles:
Hey, there's a lot going on here! I tested it out and it all looks great but I think I need a run through at some point to get my head around it all. I pushed up one commit that might help with some build issues? I was having trouble building until I added `three` as a package to the input package

cale-bradbury:
Added global node value saving, and hidden nodes so band freq/q can be nodes and live in the state/get saved to the json

package.json
```diff
@@ -2,10 +2,7 @@
   "name": "root",
   "private": true,
   "scripts": {
-    "dev": "lerna run dev --parallel --ignore @hedron/example-project --ignore @hedron/midi-test-app --ignore @hedron/clock-test-app",
-    "dev:clock": "lerna run dev --parallel --scope @hedron/clock --scope @hedron/clock-test-app",
-    "storybook": "lerna run storybook --scope @hedron/ui-core",
-    "build:engine": "lerna run build --scope @hedron/engine",
+    "dev": "lerna run dev --scope @hedron/desktop",
```
funwithtriangles, line 5: I simplified the win/mac stuff, we're all using the same thing now. This should do the same thing as `cd packages/desktop` but is a little nicer. Ideally we should also be able to add the `--include-dependencies` flag here but it breaks things when I tried it, maybe there's just too much going on for all packages to be watched at once. I also updated the readme explaining things.

funwithtriangles (approved):
I went through all the files, all makes sense, great stuff! I have made one small change to the `package.json`, hope that doesn't break things on your end. I think for now with the monorepo the most sane approach is to always do a `build` and then only run `dev` on the relevant packages for your current dev session.

PR #570, opened 2025-09-22
Server shutdown

cale-bradbury:
Shuts down the existing server before restarting it. On Windows, I meeded to restart the terminal process every time changes were made in the hedron repo,

Also adds a few filetypes to the list, and exposes the plugins on the global __HEDRON object

### DIFF
--- a/packages/desktop/src/main/SketchesServer/SketchesServer.ts
+++ b/packages/desktop/src/main/SketchesServer/SketchesServer.ts
@@ -1,18 +1,20 @@
-import path from 'path'
-import { EventEmitter } from 'events'
-import chokidar from 'chokidar'
-import { getPort } from 'get-port-please'
-import { app } from 'electron'
-import * as esbuild from 'esbuild'
-import { emptyDirSync } from 'fs-extra'
-import { createGlobalVarModuleFiles, watchWithDebounce } from './utils'
 import { getEsbuild } from '@main/getUnpackedModules'
 import { FileWatchEvents } from '@shared/Events'
+import chokidar from 'chokidar'
+import { app } from 'electron'
+import * as esbuild from 'esbuild'
+import { EventEmitter } from 'events'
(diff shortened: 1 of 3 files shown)


packages/engine/src/HedronEngine/HedronEngine.ts
```diff
@@ -64,6 +64,7 @@ export class HedronEngine {
 
   public registerPlugin(plugin: IPlugin) {
     this.plugins[plugin.id] = plugin
+    ;(window as any)['__HEDRON'].plugins = this.plugins
```
cale-bradbury: bleh, type errors due to not having this change from the other PR :( Probaby going to want to modify this to some importable function `globalHedron(): HedronGlobal` so you don't need to worry about including the types anywhere it's accessed https://github.com/nudibranchrecords/hedron/pull/568/files#diff-d0583a4070061a5125918fe96cdb1220bd1a33b0cc0157c87498c9dbfad2242f

packages/desktop/src/main/SketchesServer/SketchesServer.ts
```diff
@@ -72,6 +75,7 @@ export class SketchesServer extends EventEmitter {
         '.ply': 'text',
         '.frag': 'text',
         '.vert': 'text',
+        '.json': 'text',
```
funwithtriangles, line 78: I feel like we're going to be growing this list forever 😆

funwithtriangles (approved):
This seems like a good step forward, especially if it is helping out the developer experience, thanks! I have removed the line exposing `plugins` on the global object so I can merge this feature in, hope that's ok! 

Need to have a think about what we're doing with all this global stuff moving forward, I can see how it is very useful for quick hacks, and I'm definitely not against that! Just also wary of things getting a bit messy when it might not be so hard having a nicer API...

PR #575, opened 2025-12-20
Sketch Reorder

cale-bradbury:
Allows you to reorder sketches by moving them up/down. 

Helpful for organization, essential for experimenting with post effects without needing to destroy and recreate sketches

image
image


funwithtriangles:
Works great! Gonna merge it right away. Maybe in the future we can have a nicer drag/drop like the old Hedron but this solves the problem for now. Having drag/drop would also simplify the store a bit, can just be a 'reorderSketches' function rather than two separate 'moveSketchUp` and `moveSketchDown`. Either way this is nice and tidy as it is, thanks!

PR #578, opened 2026-01-09
Shots

funwithtriangles:
This PR adds shots, like the old Hedron. 

Sketches now have a `nodeIds` array instead of `paramIds` with nodes having a `nodeType` property with a value of either `param` or `shot`.

To fire a shot, the value for that node (in `nodeValues`) is updated with an argument object (which can just be an empty new object). This triggers a listener which fires the function.

New listeners are registered with `engine.registerShot`. Right now only sketches are registering these listeners, but any part of the app (including plugins) could do this. 

Shots can take an `shotArgs` object, which right now is not really used at the moment but opens up the potential to map incoming input data (e.g. MIDI) to specific shot arguments. MIDI input is currently adding a `_midiEvent` property to the shots arg which I'm using in a very basic way in the example `shoutout` sketch.

This PR also adds a `HedronErrorBoundary` component to `ui-core` for nicer fallbacks when a component fails (rather than whole app crashing).


packages/engine/src/store/shared/addNode.ts
```diff
@@ -2,34 +2,79 @@ import {
...
+    case 'boolean':
+      state.nodes[nodeId] = { ...baseNode, valueType, defaultValue }
+      break
+    case 'string':
+      state.nodes[nodeId] = { ...baseNode, valueType, defaultValue }
+      break
+    case 'enum':
+      state.nodes[nodeId] = { ...baseNode, valueType, defaultValue, options: config.options }
+      break
+  }
```
funwithtriangles, line 77: all of this is horrible TS related witchery. Should probably clean this up...

packages/audio-input/src/AudioInput.ts
```diff
@@ -515,6 +515,7 @@ export class AudioInput implements IPlugin {
       this.inputType,
       ({ input, optionNodes, targetNode }) => {
         if (!optionNodes.isEnabled) return
+        if (targetNode.nodeType === 'shot') return
```
cale-bradbury, line 518: eventually want to have threshold triggered shots, but not anytime soon

packages/desktop/src/renderer/components/ActiveSketch/ActiveSketch.tsx
```diff
@@ -7,13 +7,32 @@ import {
...
+import { SelectedNode } from '@components/SelectedNode/SelectedNode'
+
+interface ControlItemProps {
+  node: Node
+  sketchId: string
+}
+
+const ControlItem = ({ node, sketchId }: ControlItemProps) => {
+  const isActive = useAppStore((state) => state.selectedNodes[sketchId] === node.id)
+  const onSelectNode = useOnSelectNode(sketchId, node.id)
```
cale-bradbury, line 32: I kind of wonder if node selecting is something we want to handle at a higher (or maybe rather lower) level, such that plugins don't need to add their own rendering for the nodes options/inputs, users can expect that it will always appear in the same place, etc.

packages/desktop/src/renderer/components/GlobalDialogs/SketchModulesDialog.tsx
```diff
@@ -40,11 +40,16 @@ export const SketchCard = ({
...
+  const numShots = nodes.filter((n) => n.nodeType === 'shot').length
+
   return (
     <Card>
       <CardContent>
         <CardHeader iconName="token">{title}</CardHeader>
-        <CardDetails>Params: {params.length}</CardDetails>
+        <CardDetails>
+          Params: {numParams} • Shots: {numShots}
+        </CardDetails>
```
cale-bradbury, line 52: fine for now, but I kind of feel weird about this, eventually want skecthes to have dynamic params (beyond just initialization) and this kind of falls apart with that, plus how useful the info actually is before instancing the sketch for real (in the case of dynamic params)

packages/engine/src/HedronEngine/HedronEngine.ts
```diff
@@ -130,34 +131,82 @@ export class HedronEngine {
...
+    const unsubscribe = this.registeredShots[shotId]
+    if (unsubscribe) {
+      unsubscribe()
+      delete this.registeredShots[shotId]
+    }
+  }
+
+  public fireShot(shotId: string, value?: ShotArgsObject) {
+    // We don't directly call the shot function, instead we update the node value and let the store listener handle it
+    this.store.getState().updateNodeValue(shotId, value ?? {})
```
cale-bradbury: will this still fire if the value doesn't change?

packages/engine/src/HedronEngine/importSketchModule.ts
```diff
@@ -6,21 +6,22 @@ import {
...
 } from '@store/types'
 import { createUniqueId } from '@utils/createUniqueId'
 
 const ensureParamImported = (
   param: SketchConfigParam,
-  groupIndex: number | null,
-): SketchConfigParamImported => {
+): Omit<SketchConfigParamImported, 'groupIndex'> => {
   const valueType =
     'valueType' in param && param.valueType !== undefined ? param.valueType : 'number'
```
cale-bradbury: you were talking about explicit value types, want to log a warning if undefined?
also
```suggestion
  const valueType = param.valueType ?? 'number'
```

packages/engine/src/store/actionCreators/createReconcileSketchNodes.ts
```diff
@@ -0,0 +1,62 @@
...
+        delete state.nodeValues[oldNodeId]
+
+        // Remove vector child nodes if they exist
+        if (hasChildNodes(oldNode)) {
+          oldNode.childNodeIds.forEach((childNodeId) => {
+            delete state.nodes[childNodeId]
+            delete state.nodeValues[childNodeId]
+          })
+        }
+      }
```
cale-bradbury, line 58: thinking about this, how would you feel about adding them to some kind of temp purgatory until the app is closed, that way they can be recalled if you were just moving stuff around, or misspelled something, etc. (not in this PR, just in general)

packages/engine/src/store/shared/addNode.ts
```diff
@@ -2,34 +2,79 @@ import {
...
+        sliderMin: config.sliderMin,
+        sliderMax: config.sliderMax,
+      }
+      break
+    case 'boolean':
+      state.nodes[nodeId] = { ...baseNode, valueType, defaultValue }
+      break
+    case 'string':
+      state.nodes[nodeId] = { ...baseNode, valueType, defaultValue }
+      break
```
cale-bradbury, line 73: you can double up identical cases like this

```suggestion
    case 'boolean':
    case 'string':
      state.nodes[nodeId] = { ...baseNode, valueType, defaultValue }
      break
```

packages/engine/src/world/SketchManager.ts
```diff
@@ -13,6 +14,10 @@ type SketchUpdateParams = {
   scene: EngineScene
 }
 
+type ShotFunc = (
+  args: Omit<SketchUpdateParams, 'deltaFrame' | 'deltaTime'> & { shotArgs: ShotArgsObject },
+) => void
```
cale-bradbury: I agree that we probably don't want shots returning a params partial anymore, but it would be good to give sketches access to a hedron api they an use to update values

packages/engine/src/HedronEngine/types.ts
```diff
@@ -5,3 +5,8 @@ export type Result<T> =
 export type RendererType = 'webgl' | 'webgpu'
 
 export type CanvasSizeMode = 'fixedAspectRatio' | 'fillContainer'
+
+export type ShotArgsObject = {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  [key: string]: any
+}
```
cale-bradbury, line 12: is this what gets passed to the shot function in the sketch?

cale-bradbury (commented):
looking good :)

packages/engine/src/store/actionCreators/createAddInput.ts
```diff
@@ -16,7 +16,7 @@ export const createAddInput: SetterCreator<'addInput'> =
         const cfgImported = {
           ...cfg,
           valueType: cfg.valueType ?? 'number',
-          groupIndex: null,
+          groupIndex: 0,
           title: cfg.title ?? cfg.key,
         } as SketchConfigParamImported
```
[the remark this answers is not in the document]
funwithtriangles: Agreed this is a bit of a TS mess, needs fixing later.

packages/audio-input/src/AudioInput.ts
funwithtriangles: yep! Will need some simple latch system too, so shots can re-fire after going below the threshold. I have already done something like this with MIDI inputs, its very simple but it works

packages/desktop/src/renderer/components/ActiveSketch/ActiveSketch.tsx
funwithtriangles: I think it's good to give plugin authors freedom to display options how they like, because there are going to be loads of little edge cases where we want to show/hide certain option nodes based on some conditional logic. However I can definitely see where it would be useful to have some automatic fallback if the user doesn't define the UI for the panel, where it just grabs the option nodes and throws them in a typical grid

packages/desktop/src/renderer/components/GlobalDialogs/SketchModulesDialog.tsx
funwithtriangles: Ah I Just think it's quite a nice little bit of extra info to see while I'm browsing through sketches! Maybe we can just put something like `Params: ~{numParams}` to indicate it might not be accurate 😉

packages/engine/src/HedronEngine/HedronEngine.ts
funwithtriangles: Short answer: no, it won't fire if the value doesn't change

However, in practice, this value should _always_ be a new value. On this line, for example, if no value comes in, then `{}` evaluates to a new object every time this is called and so it will fire.

For inputs that are firing shots, they should also be always passing in some new object. Look at how we're handing it with `MidiInput.ts`:

```ts
  private handleShot: ShotHandler = ({ input, engine, midiEvent }) => {
    engine.fireShot(input.targetNodeId, { _midiEvent: midiEvent })
  }
```

The second argument here is always a newly defined object, so even if the values inside it match up perfectly, it will be evaluated as a new object every time and fire.

It's good you brought this up though, and is something we should be aware of, because some code like will only cause the shot to fire once:

```ts
  const myArgsObj = { ...whatever }
  ...
  private handleShot: ShotHandler = ({ input, engine, midiEvent }) => {
    engine.fireShot(input.targetNodeId, myArgsObj)
  }
```
`myArgsObj` is a stable reference and so the second time `handleShot` is called, the store subscriber wont fire the shot.

So to remedy this I've just committed a little change so that `fireShot` always creates a new object:

```ts
  public fireShot(shotId: string, shotArgs?: ShotArgsObject) {
    // We don't directly call the shot function, instead we update the node value and let the store listener handle it
    // We're spreading the args to create a new object reference, to ensire the store listener detects a change
    this.store.getState().updateNodeValue(shotId, shotArgs ? { ...shotArgs } : {})
  }
```

packages/engine/src/HedronEngine/importSketchModule.ts
funwithtriangles: so we've always allowed `valueType` to be undefined to try and be a bit more developer friendly, as most of the time we're dealing with numbers and so its one less line of code in config files. This is causing some stress on the typing side of things but maybe its still a price worth paying?

Thanks for this little tidy up thing, I think that weird typing was needed at some point but looks like not anymore! Nice

packages/engine/src/store/actionCreators/createReconcileSketchNodes.ts
funwithtriangles: Ah yeah that could be nice! Maybe they move over to the `AppStore` which is then stored in the user's local storage

packages/engine/src/store/shared/addNode.ts
funwithtriangles: Yeah I see the ugliness here too, unfortunately TS throws an error with your fix, because `valueType` and `defaultValue` are being cast differently based on the outcome of the switch statement. This whole file needs some love, the TS got a bit mad (partly due to me letting my AI agent go a bit wild trying to fix it...)

packages/engine/src/world/SketchManager.ts
funwithtriangles: Here it actually is passing the params to the shots! just omiting `deltaFrame` and `deltaTime` because there was no easy way for me to grab those values and pass them in (also it doesn't really make sense for shots to have these delta values anyway...)

packages/engine/src/HedronEngine/types.ts
funwithtriangles: As far as the engine is concerned a shot just has one argument, an object with any type of values. However for sketches specifically (e.g. not snapshots or whatever else) when registering the shot, we wrap the function and pass in an object with `shotArgs` as a property, as well as `params`, `scene` and whatever else might be useful in the future! I've renamed `ShotFunc` to `SketchShotFunc` to make this more clear.

PR #580, opened 2026-01-15
Video Prewarming & Clock Sync

cale-bradbury:
Video can have prewarm frames, that render at the target resolution, but do not save, helpful for letting feedback loops etc. get up and running so you can get a perfect loop.

Video rendering is consistent when using lfos, regardless of render speed or prewarmed frames.

Wanted to get this in before moving video to a plugin, as the code was done, not sure when it'll get moved to plugin, and didn't want it to get lost

### DIFF
--- a/packages/clock/src/index.ts
+++ b/packages/clock/src/index.ts
@@ -30,6 +30,7 @@ export class Clock {
   private _onNewBeat?: (beatCount: number) => void
   private _onBpmChange?: (bpm: number) => void
   private _onIsRunningChange?: (isRunning: boolean) => void
+  private _manualMode: boolean = false
 
   /**
    * @param bpm Beats per minute
@@ -62,7 +63,9 @@ export class Clock {
         this._beatCount = newBeatCount
       }
 
-      requestAnimationFrame(this.tick)
+      if (!this._manualMode) {
+        requestAnimationFrame(this.tick)
+      }
 
       this._lastTimestamp = timestamp
     }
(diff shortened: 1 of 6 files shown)


packages/clock/src/index.ts
```diff
@@ -62,7 +63,9 @@ export class Clock {
         this._beatCount = newBeatCount
       }
 
-      requestAnimationFrame(this.tick)
+      if (!this._manualMode) {
+        requestAnimationFrame(this.tick)
+      }
```
cale-bradbury, line 68: I don't love plugins having their own/tracking their own time. Think we might want to make a solution like sketches where they get updated with the time/delta from the engine, (be that by having an optional update function, or registering themselves with the engine)

packages/desktop/src/renderer/components/VideoControls/RenderTab.tsx
```diff
@@ -51,17 +53,26 @@ export function RenderTab({ renderSettings, setRenderSettings }: RenderTabProps)
       console.log = (message: string) => {
```
cale-bradbury, line 53: I don't like this and will move away from the console hijacking in the move to plugin

packages/desktop/src/renderer/utils/frameCapture.ts
```diff
@@ -132,6 +184,8 @@ window.renderFrames = async (
       console.error(`Failed to create video: ${result.error}`)
     }
   }
+
+  console.log('Render complete')
```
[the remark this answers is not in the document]
cale-bradbury: Cant turn delete the logs yet due to using it for detecting state 🤮  will be cleaned up in a future PR

packages/clock/src/index.ts
funwithtriangles: Yeah it feels weird doing this, although I will say that `requestAnimationFrame` does what it says on the tin! Plugins are guaranteed to have frame based updates by calling it

packages/desktop/src/renderer/components/VideoControls/RenderTab.tsx
funwithtriangles: ok 😅

packages/desktop/src/renderer/utils/frameCapture.ts
```diff
@@ -25,13 +26,22 @@ async function saveFrameViaIPC(
...
+  frameCount?: number,
 ): Promise<SaveFrameResponse> {
   const base64Data = dataUrl.replace(/^data:image\/png;base64,/, '')
+
+  // Apply padding if we have frameCount
+  let paddedFrameIndex = frameIndex
+  if (frameCount !== undefined && typeof frameIndex === 'number') {
+    const padding = String(frameCount - 1).length
+    paddedFrameIndex = String(frameIndex).padStart(padding, '0')
+  }
```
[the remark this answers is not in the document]
funwithtriangles: does this checkout, @cale-bradbury?

cale-bradbury:
Closed because it is handled in https://github.com/nudibranchrecords/hedron/pull/664
(no sync in that PR, but it's okay, will get that back in in it's own PR when I need it next 😆

PR #581, opened 2026-01-16
Add font loading support

cale-bradbury:
I tried writing a plugin to load unknown as url, but I just couldn't get it working 😭


funwithtriangles:
the list must grow

PR #582, opened 2026-01-16
Wireframe Solid Sketch

cale-bradbury:
Alternate solid sketch, forced wireframe but with edge thickness

No need to merge if you're not feeling it, just wanted to send it your way :)

### DIFF
--- a/packages/example-project/sketches/SolidWire/index.ts
+++ b/packages/example-project/sketches/SolidWire/index.ts
@@ -249,6 +249,7 @@ export default class Solid {
     return {
       title: 'Solid Wire',
       category: 'simple',
+      description: 'Wireframe solid geometries with adjustable edge thickness',
       params: [
         {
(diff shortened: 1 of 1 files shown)


funwithtriangles:
What's stopping us from merging the edge thickness into the main Solid sketch? 🤔

cale-bradbury:
> What's stopping us from merging the edge thickness into the main Solid sketch? 🤔

the wireframe toggle, this one doesn't support solid rendering because it's creating a mesh of the edges (which if we don't care about, then totally we can replace)

funwithtriangles:
Just tried it out, works great! And funnily enough I'm gonna need this sort of thing soon for another sketch so will definitely be copying some of this code 😅 

In terms of all these example sketches, I think for now we're fine just to pile them up but eventually I guess they'll all move to a database (so users can browse) and wont live here in this repo anyway...

PR #583, opened 2026-01-16
Midi Output & Improvements

cale-bradbury:
Adds helper functions to send midi events from Hedron (I tap into the plugin from a sketch)

Add an input option to override the value, this can be useful for using a button to set a specific float value, etc.


packages/midi-input/src/MidiInput.ts
```diff
@@ -108,6 +123,22 @@ export class MidiInput implements IPlugin {
...
+      // Apply all pending updates
+      this.pendingUpdates.forEach((value, nodeId) => {
+        storeState.updateNodeValue(nodeId, value)
+      })
+
+      // Clear pending updates
+      this.pendingUpdates.clear()
+      this.updateScheduled = false
+    })
+  }
```
funwithtriangles: This is great but I'm wondering if this should happen at a store level rather than at this MIDI level? Maybe `updateNodeValue` should be doing the batching instead of this MIDI plugin, so all areas of the app benefit. Not sure exactly how this would be tackled though...

funwithtriangles: Can you send me an example of where you were experiencing performance issues? I'm not entirely convinced that multiple store updates is the real issue. Could have been some weird React component thing going on and this fix didn't really do much (e.g. I've had some issues and then a refresh just cleared them). Batching updates is probably still a good idea but I'd only expect minor performance gains... I've tried replicating the issue on `alpha` without much luck.

cale-bradbury: yeah I think it depends on midi device going off spec (they are supposed to limit messages to one per x ms, but maybe some vendors are only careful with that for din instead of usb). I'll try and replicate on alpha and see if I can get some video.

funwithtriangles: ah right. makes sense! Maybe I should give it a go with my other device and see if i can get some more chaos to happen

funwithtriangles: I've merged in your other approach, do you wanna revert this now?

packages/midi-manager/src/MidiManager.ts
```diff
@@ -181,6 +188,7 @@ export class MidiManager {
     this.learnPromise = new Promise((resolve) => {
       this.learnResolve = resolve
       this.learnListener = (event: MIDIEvent) => {
+        if (event.type === MidiMessageType.Clock) return
```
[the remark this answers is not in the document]
cale-bradbury: we may want to learn start/stop/continue

funwithtriangles:
All looks good! I didn't test the output stuff but did test the override and it works 👍

PR #584, opened 2026-01-22
Buffer store values to only update once per frame

cale-bradbury:
Works with my launch control XL, for which quickly changing values causes a fps drop on alpha


funwithtriangles:
nice! This generally feels like a great step forward and a nice clean approach, thanks! I've tested it and nothing exploded 😅

PR #587, opened 2026-01-22
Apps and packages

funwithtriangles:
Moves all app-like packages into a new directory, `apps`. This keeps the separation clear between apps and the packages they consume. I've also added LICENSE files, with an explanation in the readme of what they mean.

Also used this opportunity to publish on npm, had to wiggle some config files to get it to work


funwithtriangles:
@cale-bradbury you cool with this? Also see if you agree with my license structure. Very strong license only for the desktop app itself, the rest is basically "do what you want"

cale-bradbury (commented):
Looks good, but I'll give it a windows test over the weekend

funwithtriangles:
@cale-bradbury reckon you could describe the build issue you're having here? maybe we can narrow it down

PR #595, opened 2026-01-26
Gamepad Input

cale-bradbury:
Adds gamepad input, featuring controller index abstraction, support for axis & buttons, input learning.

The main assumption this makes is limiting to 4 abstracted inputs (theoretically infinite true inputs, overlapping their targets. ie 8 switch joycons can be mapped down to the 4 controllers they effectively behave as, or having one controller in your 'booth' one out and away for the crowd to play with mapped the virtual index so you can engage when no one else is playing, etc.)

Up for review of core system, and can be merged whenever you would like, I'll slowly be chipping away at the todos, in this PR or the next

image

todo:
- [x] axis smoothing
- [x] axis normalization (they normally don't actually report as perfect circles, might need normalization, might need capping)
- [x] 2 axis as angle
- [x] 2 axis as distance
- [x] isEnabled toggle
- [x] axis button blocking option (ie hold x then the axis data is applied to the param, could be skipped if we could get recursive node inputs)
- [x] button on/off value
- [x] button toggle
- [ ] accelerometer support
- [ ] debug vibration (to tell which is which when multiple with the same name are connected)
- [ ] input list link to node
- [ ] input list enable toggle


packages/gamepad-input/src/GamepadManager.ts
```diff
@@ -0,0 +1,235 @@
...
+        else if (!isPressed && wasPressed) {
+          this.emitEvent({
+            controllerIndex: logicalIndex,
+            index: buttonIndex,
+            inputType: GamepadInputType.Button,
+            value: 0,
+            isPressed: false,
+          })
+        }
+      })
```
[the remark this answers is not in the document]
cale-bradbury: I think analog buttons are treated as axis? But I'll need to source a controller to test this with, could be patched in later

packages/gamepad-input/src/GamepadInput.ts
```diff
@@ -0,0 +1,627 @@
...
+    if (distance < deadZone) {
+      return 0.5 // Return to center
+    }
+
+    // Remap from [deadZone, cap] to [0, 1] and clamp
+    const remapped = (distance - deadZone) / (cap - deadZone)
+    const clamped = Math.min(1, Math.max(0, remapped))
+
+    // Convert back to [0, 1] range
+    return 0.5 + (clamped * sign) / 2
```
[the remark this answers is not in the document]
cale-bradbury: I don't know how much I care here, use at your own risk, don't think we have a nice way of updating the slider min/max, and would be weird if the shown value didn't reflect the actual, use at your own risk

packages/gamepad-input/src/GamepadInput.ts
```diff
@@ -0,0 +1,627 @@
...
+              (storeState.nodeValues[`${input.targetNodeId}-sliderMin`] as number) ?? 0
+            const sliderMax =
+              (storeState.nodeValues[`${input.targetNodeId}-sliderMax`] as number) ?? 1
+            const scaledValue = combinedValue * (sliderMax - sliderMin) + sliderMin
+
+            this.targetValues.set(input.id, scaledValue)
+          }
+
+          // For 2-axis mode, we handle the update above, so return early
+          return
```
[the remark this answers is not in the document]
cale-bradbury: I'm fine leaving this for now, I think we need to figure out some shared layer for converting from the input value, to the true value, to stop each input needing to re-write the same logic (and potentially handle it differently)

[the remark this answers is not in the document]
cale-bradbury: Like I'm thinking eventually we will want gradient support for colors, sending through a numeric value from the input, but don't want every input class to need to track the additional colors/LUT/.gradient/etc. as that logic would be best shared, it just sends a numeric value and the rest is handled internally

packages/gamepad-input/src/GamepadInput.ts
```diff
@@ -0,0 +1,357 @@
...
+   */
+  public handlers: {
+    handleShot: ShotHandler
+    enum: ValueHandler<NodeParamEnum>
+    boolean: ValueHandler<NodeParamBoolean>
+    number: ValueHandler<NodeParamNumber>
+    string: ValueHandler<NodeParamString>
+    rgb: ValueHandler<NodeParamRGB>
+    vector3: ValueHandler<NodeParamVector3>
+  }
```
funwithtriangles, line 70: I got runtime errors when trying to use a vector2 with gamepad, but no TS errors. Either we want there to be TS errors when a new value type has been defined or we have some general catch for unknown value types.

You could type `handlers` like this, to catch all param types, forcing the property to include all value types. But I'm thinking it makes more sense to handle this by just only having the supported types and catching unknown/unsupported types in a single place.

```ts
  public handlers: {
    handleShot: ShotHandler
  } & { [K in Exclude<NodeValueType, null>]: ValueHandler<Extract<Param, { valueType: K }>> }
```

packages/gamepad-input/src/GamepadInput.ts
```diff
@@ -0,0 +1,357 @@
...
+    })
+
+    this.handlers = {
+      handleShot: baseHandlers.handleShot,
+      enum: baseHandlers.handleEnum,
+      boolean: baseHandlers.handleBoolean,
+      number: baseHandlers.handleNumber,
+      string: baseHandlers.handleUnsupportedString,
+      rgb: baseHandlers.handleUnsupportedRGB,
+      vector3: baseHandlers.handleUnsupportedVector3,
```
funwithtriangles, line 91: feels weird to need to specify all of these unsupported ones, and we'll need an extra `vector2` in there now as well.

packages/gamepad-input/src/GamepadHandlers.ts
```diff
@@ -0,0 +1,274 @@
...
+    input,
+    targetNode,
+    gamepadEvent,
+  }) => {
+    console.warn(
+      `Gamepad Input: Unsupported value type for node ${input.targetNodeId}. Value: ${gamepadEvent.value}, Type: ${targetNode.valueType}`,
+    )
+
+    return null
+  }
```
funwithtriangles, line 279: As mentioned elsewhere, feels weird to have repeated unsupported handlers here, can't we just have a single generic one? Would also be more futureproof as new value types come in.

funwithtriangles (approved):
Wowzer, this is so much fun! Amazing stuff, feels like magic! Only one minor thing I've noticed and I wouldn't even let it stop merging if you don't feel like fixing it, I can handle it at some point if I feel the need (I think this pattern needs looking at across all the inputs tbh). Let me know if ya wanna deal with it, otherwise I'll just hit the button.

I love how few files this touches outside of its own package! 😎

packages/gamepad-input/src/GamepadInput.ts
cale-bradbury: in theory eventually they will be supported, just not in the first pass, because maybe we want some kind of shared handling across input packages

packages/gamepad-input/src/GamepadHandlers.ts
cale-bradbury: yeah, placeholders, but I agree, probably not worth it

packages/gamepad-input/src/GamepadInput.ts
funwithtriangles: ok I was wrong, we did get TS errors related to the missing `vector2`. I just went ahead and fixed it following the same pattern you have going.

packages/gamepad-input/src/GamepadInput.ts
funwithtriangles: yeah, or we have a notion of "Supported value types" so the option to add certain inputs doesn't even appear for those nodes

packages/gamepad-input/src/GamepadInput.ts
cale-bradbury: yeah, that looks good, but maybe gets tricky if we end up with non-nudibranch managed packages that don't live in the mono repo, they might slip out of date with what the type actually is and end up with runtime errors anyways. That said, probably a good move, to add this in, but would be nice if it lived in types.ts and just became `public handlers: InputHandlers`

PR #613, opened 2026-02-17
Vector 2 value type with canvas scrub

funwithtriangles:
Todo:

- [x] `useElementScrub` on both axes
- [x] tidy viewer component (make a hook)
- [x] viewer component should have some active style when scrubbable
- [x] pointer should change to XY thingy
- [ ] node option to disable/enable scrubbing on canvas
- [x] test with proper 2D usage
- [x] map to `min` and `max` of the node
- [x] tidy up `SingleSlider` for vec2 and vec3, lots can be shared here. Handle slider min and max


apps/desktop/src/renderer/components/Viewer/index.tsx
```diff
@@ -1,13 +1,46 @@
-import { useCallback } from 'react'
+import { useCallback, useRef } from 'react'
+import { useElementScrub } from '@hedron-gl/ui-core'
+
 import c from './styles.module.css'
+import { useSelectedNode } from '@components/hooks/useSelectedNode'
 import { engine } from '@renderer/engine'
 
 export const Viewer = (): JSX.Element => {
+  const selectedNode = useSelectedNode()
```
cale-bradbury: wondering if this could be good as it's own file that is passed in a html element, that way the same base logic could be used for a little x/y pad, potentially as the param view, maybe you could spawn multiple x/y pads that float over the preview canvas, etc.

funwithtriangles: Yeah definitely would make sense as its own component at some point! Although `useElementScrub` is doing all the heavy lifting which is very reusable

funwithtriangles: Eventually I'd have vector2 params pop out a little XY pad.  (e.g. https://tweakpane.github.io/docs/input-bindings/#point)

PR #614, opened 2026-02-17
Sketch Search

cale-bradbury:
When opening the add sketch popup, a search field is focused, filters sketches as you type, pressing enter will add the top result.

image

### DIFF
--- a/apps/desktop/src/main/menu.ts
+++ b/apps/desktop/src/main/menu.ts
@@ -66,6 +66,18 @@ const template: Electron.MenuItemConstructorOptions[] = [
   projectMenu,
   displayMenu,
   {
+    label: 'Edit',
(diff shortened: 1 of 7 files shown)


funwithtriangles:
Nice stuff! Thanks! I made a few changes:
- changed the input type to `search` instead of `text`. This gives you that `Escape` functionality by default and adds a little "X" button to clear the search too
- Did some electron configuring so we can use CMD+A to select all (this also fixes things in other areas of the app)
- Moved the search filtering functionality into a hook and did a few little tweaks to the code to clean it up
- Moved `SketchCard` into its own file
- styled the input to be inline with the rest of the app

We should probably have a component for a basic text/search input like this, I'll get round to it when we start repeating ourselves!

PR #615, opened 2026-02-17
Slider min max any order

cale-bradbury:
Before if max was less than min, you would not be able to drag the slider, now it works regardless of which one is actually bigger.

Min/max more so means 0/left edge and 1/right edge, not necessarily which one is actually bigger

image


funwithtriangles:
Great work, some fiddly logic there too!

PR #616, opened 2026-02-17
Resizable left/right half

cale-bradbury:
Adds a handle that can be dragged to make either side larger.

Feel free to just close this PR if you want to hold off and tackle this in more of a window/tab layout system kind of way :)

image


funwithtriangles:
Lovely! This is really good and I've also been craving it. I'll do a little bit of tidying on this very soon before merging it. We have some nice stuff coming with #613 that will help too. It would also make sense to store the ratio value in the app store, so per project it is saved.

funwithtriangles:
I moved the dragging logic into a hook, `useHandleDrag` and tweaked the style. It is visible thinner now (and no gradient) but the hitbox is actually larger than before (using a pseudo element). Also making use of `setGlobalCursor` util which I just introduced with #613 

For now I'm not gonna bother saving this ratio value in the app store, it doesn't take any effort to resize this when you need it and no need to add more complexity to the codebase or this PR!

Weeee! I'm happy to have this one in, thanks again!

PR #617, opened 2026-02-18
Tab submits number change, Escape reverts

cale-bradbury:
I often want to just hit tab to cycle through and rapidly change a bunch of values, I think this pretty standard behaviour for inputs. Escape can be used if you want to undo your change


funwithtriangles:
Works great, nice addition! Thanks!

PR #618, opened 2026-02-18
Save the active sketch to the project file, open that sketch when loaded

cale-bradbury:
Saw the code comment about it breaking, but seems to work fine, maybe that was from the old version? I'll let you know when I test on windows


funwithtriangles:
Hah! Thanks for fixing this. This exact lack of functionality was annoying me recently, I have no memory of writing that comment, maybe it was meant to be fixed before merging and I forgot 🙈 

Seems to work for me 👍

PR #620, opened 2026-02-20
safely try sketch imports

funwithtriangles:
This PR does a few things:
- Stops app from crashing when some import fails, now we'll just get a console error
- Don't show any errors or warnings for folders in the sketches dir that don't have an `index.js` or `index.ts` (e.g. empty directories are fine)
- No longer watch unrelated files (e.g. `mySketch/utils/index.js`)
- No longer see 404 errors for missing configs if we are using `getConfig`
- Allows for sketches without any config file (will still show 404 error in console)
- ⚠️  **Breaking:** Requires that the `getConfig` method of sketches is `static`

@cale-bradbury can you check if this is working on Windows? Also how do you feel about `getConfig` being a `static` method on sketches? Is there ever a time when one of your sketches is doing something more clever with `getConfig` and it can't be a static method? There is a workaround for this but it does have a few flaws.


packages/engine/src/HedronEngine/importSketchModule.ts
```diff
@@ -135,37 +136,42 @@ export const importSketchModule = async (
...
     let config: SketchConfigImported
-    if ((await fetch(configPath)).status !== 200) {
-      // No config file found
-      // Try instancing the sketch, and call getConfig() on it
-      const tempModule = new module(undefined)
-      const unprocessedConfig = tempModule.getConfig?.() as SketchConfigRaw | undefined
 
-      tempModule.dispose?.()
+    // First try to get config from static getConfig()
+    const unprocessedConfigFromGetConfig = module.getConfig?.() as SketchConfigRaw | undefined
```
funwithtriangles, line 157: @cale-bradbury  going for the static method here means we don't have to worry about instantiating a module temporarily. An issue I was seeing is that if I had a sketch that expected the constructor to have arguments, I was seeing console errors where it couldn't find them. I know this is all temporary and so it's not a huge deal but it felt a bit flaky. If this really doesn't work for you I can revert back to wrapping this in `try/catch` and instantiating, as it was before.

packages/engine/src/HedronEngine/importSketchModule.ts
```diff
@@ -135,37 +136,42 @@ export const importSketchModule = async (
 
     // Get the sketch module
     const sketchPath = `${baseUrl}/${moduleId}/index.js?${cacheBust}`
-    if ((await fetch(sketchPath)).status !== 200) {
-      return Promise.reject(`Sketch module not found: ${sketchPath}`)
```
funwithtriangles, line 139: this Promise.reject was handled as uncaught and therefore crashing the app. Now we just return result with the `success` flag as `false` and an error message, letting the app handle it further along.

packages/engine/src/HedronEngine/importSketchModule.ts
cale-bradbury: Oh I was specifically using the lack of engine as a check to see if I needed to actually instance the rest of the sketch, an early exit for ones doing a bunch of loading/processing that isn't needed for the config

cale-bradbury (approved):
Windows is working well, and the error handling is very nice :)

PR #626, opened 2026-03-01
Midi Global Panel & Value Smoothing

cale-bradbury:
Adds a panel where you can see incoming midi notes.
Adds a global midi option for smoothing

image

The history looks a little messy/glitched, so feel free to squash and merge when the time comes.

Not much here yet, but may as well get it in as some boilerplate to build off of (plus the smoothing is nice)(


packages/midi-manager/src/MidiManager.ts
```diff
@@ -235,6 +265,64 @@ export class MidiManager {
...
+    current: number,
+    target: number,
+    callback: (value: number) => void,
+  ): void {
+    this.smoothedValues.set(key, {
+      current,
+      target,
+      callback,
+    })
+  }
```
funwithtriangles, line 277: I simplified the logic here by just handing in a current value from outside. The main reason for doing this is that now we'll always have smoothed behaviour, from a cold start too. I was getting some crunchy looking changes when doing small increments, this fixed that.

packages/midi-manager/src/MidiManager.ts
```diff
@@ -46,13 +46,38 @@ export type MIDIEvent = {
...
+   * Smoothing factor for MIDI values (0 = no smoothing, closer to 1 = more smoothing)
+   */
+  public get smoothing(): number {
+    return this._smoothing
+  }
+
+  public set smoothing(value: number) {
+    // Never allow smoothing of 1, as that would cause values to never update
+    this._smoothing = Math.max(0, Math.min(0.99, value))
+  }
```
funwithtriangles: Clamping smoothing because at 1, no updates happen, which I'd imagine we'd never want. Also renamed `SMOOTHING` to `smoothing` because the all caps felt weird as a property!

packages/midi-manager/src/MidiManager.ts
```diff
@@ -46,13 +46,38 @@ export type MIDIEvent = {
...
+type SmoothingEntry = {
+  target: number
+  current: number
+  callback: (value: number) => void
+}
+
+/**
+ * Threshold for stopping smoothing when close enough to target
+ */
+const SMOOTHING_THRESHOLD = 0.001
```
funwithtriangles, line 58: Move this outside of the class because its a constant, just made more sense to me 🤷

funwithtriangles (approved):
Hey, this is great and something I'll definitely feel the benefit from! I noticed that the smoothing wasn't perfect:

If I started updating a value very slowly, I was still getting crunchy behaviour, because the first tick had `target` and `current` set the same, so we'd always get an initial first step-like update. I fixed this by handing `current` in from the outside. This also had the benefit of making `setSmoothedValue` a much simpler function!

I also did a couple of other little changes, see my comments 👼 

I hope that's all cool with you, gonna keep this open in case you disagree or I've broken something!

packages/midi-input/src/MidiInput.ts
```diff
@@ -129,12 +139,31 @@ export class MidiInput implements IPlugin {
...
 
-    const value = this.getValue(optionNodes, midiEvent)
-    return (value / 127) * (sliderMax - sliderMin) + sliderMin
+    const targetValue =
+      (this.getValue(optionNodes, midiEvent) / 127) * (sliderMax - sliderMin) + sliderMin
+
+    // Use MidiManager's smoothing system
+    this.midiManager.setSmoothedValue(
+      input.id,
+      currVal as number,
```
funwithtriangles, line 158: Now handing in the node's current value to the smoothing function

packages/midi-manager/src/MidiManager.ts
cale-bradbury: hell yeah, nice one

packages/midi-manager/src/MidiManager.ts
cale-bradbury: I've used 1 as a way to freeze, feel like this would be better if it came from the params min/max, which I would feel good about having this as the default for :)

funwithtriangles: Agreed, no need for the extra complexity here! We can add the min/max default params later if it feels necessary.

funwithtriangles: (I have reverted this change)

PR #627, opened 2026-03-01
Camera is now private with a get/set

cale-bradbury:
Makes camera a get/set, so that way it can be replaced by sketches (use at your own risk, maybe we move this to it's own plugin at some point).

Needed this to switch between Perspective and Orthographic


funwithtriangles:
Very useful, thanks! Eventually I guess we have a more unified API for all of these little things, but this is a nice and easy solution for now.

PR #631, opened 2026-03-19
Inputs on inputs

funwithtriangles:
- Node control layout is always now vertical
- Node controls now display input count
- Can now open any node in the selected node panel
- Selected node for sketch now has breadcrumbs for navigating back up

@cale-bradbury you cool with the design change with everything being vertical like this? It made sense to squeeze that input count in and has the benefit of wider sliders, but I'm not entirely sure I love it 😅 We can always change it later...

https://github.com/user-attachments/assets/d9f2f0c5-d67b-4b99-8be2-6f1ceb8b3752

### DIFF
--- a/packages/ui-core/src/stories/NodeControl.stories.tsx
+++ b/packages/ui-core/src/stories/NodeControl.stories.tsx
@@ -8,6 +8,8 @@ import {
   NodeControlMain,
   NodeControlTitle,
   NodeControlInner,
+  NodeControlInfo,
+  NodeControlInputCount,
 } from '@components/NodeControl/NodeControl'
 
 import { ControlGrid } from '@components/ControlGrid/ControlGrid'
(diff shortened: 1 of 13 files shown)


cale-bradbury:
into the vertical change!

packages/ui-core/src/stories/NodeControl.stories.tsx
```diff
@@ -342,3 +275,65 @@ export const NumberReversed = ({ title = 'Short Name', isActive, layout, onClick
     </NodeControl>
   )
 }
+
+export const NumberWithInputCount = ({ title = 'Brightness', isActive, onClick }: BasicProps) => {
+  const ref = useRef<FloatSliderHandle>(null)
+
+  useEffect(() => {
+    ref.current?.updateValue(Math.random() * 1)
```
cale-bradbury: whats up with the `* 1`?

cale-bradbury (approved):
Looking good!

packages/ui-core/src/stories/NodeControl.stories.tsx
funwithtriangles: Haha this is so silly 😆 I think it snuck in there in other stories somehow and then when getting AI to generate more similar stories it just copied the pattern across. I'll tidy it up.

PR #635, opened 2026-03-28
Open sketch source

cale-bradbury:
Adds a dropdown option to help you quickly open the source index.ts/js for a sketch in it's default program

image


apps/desktop/src/main/handlers/openSketchSourceFile.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @returns Promise<{ success: boolean; error?: string }>
+ */
+export const openSketchSourceFile = async (
+  sketchesDir: string,
+  moduleId: string,
+): Promise<{ success: boolean; error?: string }> => {
+  try {
+    const tsPath = path.join(sketchesDir, moduleId, 'index.ts')
+    const jsPath = path.join(sketchesDir, moduleId, 'index.js')
+    let filePath = tsPath
```
cale-bradbury, line 18: instead of all this (regenerating the path), do we want to just cache the path on load @funwithtriangles? might be handy if we ever support sketch root files with different names

apps/desktop/src/main/index.ts
```diff
@@ -2,6 +2,7 @@ import { app, BrowserWindow, dialog, ipcMain, screen, session } from 'electron'
 import { electronApp, optimizer } from '@electron-toolkit/utils'
 import { REDUX_DEVTOOLS, installExtension } from '@tomjs/electron-devtools-installer'
 import { ProjectData } from '@hedron-gl/app-store'
+import { openSketchSourceFile } from './handlers/openSketchSourceFile'
```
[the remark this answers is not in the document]
funwithtriangles: ignore this one @cale-bradbury. Our eslint rules dictate this sort of thing. They disallow going back up the tree (e.g. `../../foo`) but this way round is fine.

apps/desktop/src/main/handlers/openSketchSourceFile.ts
funwithtriangles: let's discuss today, don't have my head in this one right now

PR #660, opened 2026-07-09
Timeline: fire shots from keyframes

cale-bradbury:
- Keyframe.valueType widens to admit 'shot' alongside ParamValueType, since shots aren't params and their keyframes are momentary triggers rather than held values.
- TimelineManager: shot tracks are edge-detected via a per-track count of keyframes crossed (time <= position), separate from the existing held-value logic used for every other type. A shot fires once per forward crossing regardless of scrub direction or jump size, and never double-fires when sitting past a keyframe.
- useTimelineHandlers: inserting a keyframe on a shot-targeted track stores a constant `value: true` marker instead of reading a (nonexistent) param value.

### DIFF
--- a/packages/timeline/src/TimelineInput.ts
+++ b/packages/timeline/src/TimelineInput.ts
@@ -7,6 +7,7 @@ import {
   OptionNodesFromConfigs,
   isParamVector,
   ParamNode,
+  ParamValue,
   ShotNode,
   isParamVectorComponent,
 } from '@hedron-gl/engine'
@@ -121,14 +122,29 @@ export class TimelineInput implements IPlugin {
           }
 
           const tracks = getKeyframeTracks(allTracks)
-          const changedTargetNodeIds = changedTrackIds.map(
-            (trackId) => tracks.find((track) => track.id === trackId)?.targetNodeId ?? trackId,
-          )
 
-          engine.setMultipleParamValues(
-            changedTargetNodeIds,
-            changedTrackIds.map((trackId) => changed[trackId]),
-          )
+          const paramTargetNodeIds: string[] = []
+          const paramValues: ParamValue[] = []
+
+          for (const trackId of changedTrackIds) {
+            const targetNodeId = tracks.find((track) => track.id === trackId)?.targetNodeId ?? trackId
+            const targetNode = engine.getNode(targetNodeId)
+            const value = changed[trackId]
+
+            if (targetNode?.nodeType === 'shot') {
+              if (value) {
+                engine.fireShot(targetNodeId)
+              }
+              continue
+            }
+
+            paramTargetNodeIds.push(targetNodeId)
+            paramValues.push(value)
+          }
+
+          if (paramTargetNodeIds.length > 0) {
+            engine.setMultipleParamValues(paramTargetNodeIds, paramValues)
+          }
         }
       })
     })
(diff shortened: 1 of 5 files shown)


packages/timeline/src/components/TimelineGlobalPanel/useTimelineHandlers.ts
```diff
@@ -50,25 +50,31 @@ export const useTimelineHandlers = ({ engine, manager }: UseTimelineHandlersPara
...
-
-      const keyframe = {
-        id: crypto.randomUUID(),
-        time,
-        valueType: targetParam.valueType,
-        value: targetParamValue,
+      let keyframe: Keyframe
+      if (targetNode.nodeType === 'shot') {
+        // Shot keyframes are momentary triggers; the value isn't used to decide firing.
+        keyframe = { id: crypto.randomUUID(), time, valueType: 'shot', value: true }
```
cale-bradbury: I feel like at some point we will want shot to be a param, in the sense of being able to have function args maybe as the value. in the meantime, we branch

funwithtriangles: Shots can already pass a `ShotArgsObject` when firing, we're just not using it in many places. The midi input passes on the midi event so that sketch shots can use the note information. So maybe maybe shot keyframes could hand over special values for each keyframe?

cale-bradbury: would need the UI, but nice to know it's in place

packages/timeline/src/TimelineInput.ts
```diff
@@ -121,14 +122,28 @@ export class TimelineInput implements IPlugin {
...
+          for (const trackId of changedTrackIds) {
+            const targetNodeId =
+              tracks.find((track) => track.id === trackId)?.targetNodeId ?? trackId
+            const targetNode = engine.getNode(targetNodeId)
+            const value = changed[trackId]
+
+            if (targetNode?.nodeType === 'shot') {
+              engine.fireShot(targetNodeId)
+              continue
+            }
```
funwithtriangles, line 138: Rather than `valueType` of `shot`, I've added `nodeType` to keyframes, feels more robust. Also track values for shots will always just be `true`, so no need to check what the value is

packages/timeline/src/TimelineManager.ts
```diff
@@ -69,6 +71,24 @@ export class TimelineManager {
...
 
+  // Shot keyframes are momentary triggers, not held state: fire once whenever the number of
+  // keyframes crossed (time <= position) increases since the last check.
+  private getShotFired(track: TimelineManagerTrack): boolean {
+    const sorted = this.sortedKeyframesCache.get(track.id) ?? []
+    const count = sorted.filter((kf) => kf.time <= this.position).length
+    const lastCount = this.shotKeyframeCount.get(track.id) ?? 0
+    this.shotKeyframeCount.set(track.id, count)
+    return count > lastCount
+  }
```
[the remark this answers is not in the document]
funwithtriangles: let's leave this for now, if we noticed this bug later we can fix it later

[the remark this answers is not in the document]
funwithtriangles: ok i solved this simply by changing `<=` to `<`

funwithtriangles:
Thanks for this! I reworked this slightly so that we have `nodeType` on keyframes, rather than hijacking `valueType` for shots. Shot keyframes now have no value associated with them. Later we might want to include the `ShotArgsObject` with the keyframe but I still think it makes sense to have distinct keyframe types here, as they are treated quite differently.

PR #661, opened 2026-07-09
Timeline: lerp for numeric keyframes, fix keyboard shortcuts stealing focus

cale-bradbury:
- TimelineManager now interpolates between two keyframes for 'number' params instead of holding until the next one is reached. Discriminated by valueType (not typeof value) so numeric-backed enums correctly keep holding discretely instead of interpolating between option indices.
- Timeline's global keydown handler (i/x shortcuts) now ignores events when focus is on an input/textarea/contenteditable element, so typing "i" into a string param's text field no longer inserts a keyframe.

### DIFF
--- a/packages/timeline/src/TimelineManager.ts
+++ b/packages/timeline/src/TimelineManager.ts
@@ -66,6 +66,22 @@ export class TimelineManager {
       lastIndex = i + 1
     }
     this.lastKeyframeIndex.set(track.id, lastIndex)
+
+    // 'number' params interpolate towards the next keyframe rather than holding until it's
+    // reached. Checked via valueType (not typeof value) since enum values can also be numbers,
+    // but represent discrete choices that shouldn't be interpolated between.
+    const current = sorted[lastIndex - 1]
+    const next = sorted[lastIndex]
+    if (
+      current?.valueType === 'number' &&
+      next &&
+      typeof current.value === 'number' &&
+      typeof next.value === 'number'
+    ) {
+      const t = (this.position - current.time) / (next.time - current.time)
+      return current.value + (next.value - current.value) * t
+    }
+
     return value
   }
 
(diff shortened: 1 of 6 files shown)


packages/timeline/src/TimelineManager.ts
```diff
@@ -66,6 +66,22 @@ export class TimelineManager {
...
+    // 'number' params interpolate towards the next keyframe rather than holding until it's
+    // reached. Checked via valueType (not typeof value) since enum values can also be numbers,
+    // but represent discrete choices that shouldn't be interpolated between.
+    const current = sorted[lastIndex - 1]
+    const next = sorted[lastIndex]
+    if (
+      current?.valueType === 'number' &&
+      next &&
+      typeof current.value === 'number' &&
+      typeof next.value === 'number'
```
cale-bradbury:
```suggestion
      next
```

it's probably overkill?

packages/timeline/src/components/Timeline/Timeline.tsx
```diff
@@ -66,6 +66,10 @@ export function Timeline({
     }
 
     const handleKeyDown = (e: KeyboardEvent) => {
+      const target = e.target as HTMLElement | null
+      if (target?.tagName === 'INPUT' || target?.tagName === 'TEXTAREA' || target?.isContentEditable) {
+        return
+      }
```
funwithtriangles, line 72: Nice simple solution, works for now! Maybe later we want to do something more explicit, like have some notion of being focused on the timeline

packages/timeline/src/TimelineManager.ts
funwithtriangles: Yeah, I tightened up the keyframe types so that this becomes inferred based on the `valueType` property of the keyframe, so no need for this `typeof` stuff!

funwithtriangles (approved):
Great stuff, I did a bit of type work on top so we don't need as many guards when checking keyframe values

PR #663, opened 2026-07-09
Timeline: Input display and auto track selection

cale-bradbury:
When creating a new timeline input, or selecting a param with a timeline track, auto-select that params track.

Display a single lane track in the input view.

image


packages/timeline/src/components/Timeline/Timeline.tsx
```diff
@@ -1,55 +1,69 @@
...
+  const fallbackId = useId()
+  const componentId = _componentId ?? fallbackId
+
+  const isActiveComponent = componentId === activeTimelineComponentId
+
+  const [_selectedKeyframes, _setSelectedKeyframes] = useState<string[] | null>(null)
+
+  const selectedKeyframes = isActiveComponent ? _selectedKeyframes : null
+  const selectedTrackId = isActiveComponent ? _selectedTrackId : null
+
```
funwithtriangles, line 50: this means that only once instance of the timeline component can have a selected track, therefore only allowing for the keyboard events to fire in one place at a time

packages/app-store/src/appStore.ts
```diff
@@ -66,6 +67,15 @@ export interface AppState {
...
+
+  /** @deprecated -- this state handler will be moved to the timeline plugin */
+  setActiveTimelineComponentId: (id: string | null) => void
+  /** @deprecated -- this state will be moved to the timeline plugin */
+  activeTimelineComponentId: string | null
+  /** @deprecated -- this state handler will be moved to the timeline plugin */
+  setSelectedTimelineTrackId: (trackId: string | null) => void
+  /** @deprecated -- this state will be moved to the timeline plugin */
+  selectedTimelineTrackId: string | null
 }
```
funwithtriangles, line 79: We'll need to move this into some state management that lives with the plugin somehow

PR #664, opened 2026-07-10
Timeline Video Render

cale-bradbury:
This will likely have some conflicts after the other PRs, I'll clean this up once those are merged.

Moves video rendring into a plugin that uses values from the timeline.
Moves timing control to global hedron update loop instead of per-plugin `requestAnimationFrame`
Cleaned up hacky parts of the existing video code, watching for signals in console.logs, calling functions from `window`


funwithtriangles:
So I've tested this and I love how easy it is to immediately get a video/png-sequence! However I have found a few bugs:

- Rendering with width and height set to 0 (e.g. use the canvas size) is causing ffmpeg errors (`width not divisible by 2 (701x394)`), maybe its better to just force an output size?
- When I set the width and height I get a runtime error: `Renderer not set`

Another thing I noticed:
- having the export config as params/shots doesn't really make sense to me. I can see the convenience of doing it this way but we probably should have some non-node way of doing stuff like this. Let me know if you need some help with this, I can have a think about it. Seeing a shot button for "browse" did make me chuckle though 😆. If it's too much work I don't mind keeping it this way for now and we can clean it up later on.

cale-bradbury:
> * Rendering with width and height set to 0 (e.g. use the canvas size) is causing ffmpeg errors (`width not divisible by 2 (701x394)`), maybe its better to just force an output size?

Yeah, need to have it round up to the nearest even number maybe, I'm also not opposed to a forced size.

> * When I set the width and height I get a runtime error: `Renderer not set`

uhoh, have not  seen that one myself

> * having the export config as params/shots doesn't really make sense to me. I can see the convenience of doing it this way but we probably should have some non-node way of doing stuff like this. Let me know if you need some help with this, I can have a think about it. Seeing a shot button for "browse" did make me chuckle though 😆. If it's too much work I don't mind keeping it this way for now and we can clean it up later on.

The main reason for params was to get it saved in the project file, don't remember if we have a way to save arbitrary data, maybe we don't show the nodes ui and pass through from a text box, or maybe we can add a filter to a string param to enforce text only (maybe we already have this?)

PR #666, opened 2026-07-10
Midi Auto-Learn & On+Off Mode

cale-bradbury:
- Toggle to automatically go into midi learn mode when a new input is added.
- On/Off mode that is a value when pressed, but returns to 0 when released (value when held)


funwithtriangles:
Thanks for this, I did a fair amount of cleanup:
- using `getOptionNodes` (new)  and/or `useOptionNodes` where applicable
- added an extra option node of `noteMode` (on, off, on/off) and reduced the midi message `type` options to simple `controlChange` or `note`.  So now we can relearn without changing the note mode itself. Feels better to have these behaviours separate
- pulled out the logic for matching midi events with the input option nodes into `doesMidiEventMatchInput`
- reverted the special case where `noteMode` is `on/off`. It feels like this might as well cycle the enum on both on/off events, rather than the guard we had, otherwise the behaviour matches exactly with `on` mode
- extra component optimisation selector stuff (useParamValue)
- added constants and utils files for organisation

PR #671, opened 2026-07-17
engine.renderer.requestFrame

funwithtriangles:
While working with the output window always open, I was noticing some stuttering in the output window, which wasn't happening in the preview canvas. It definitely got worse for more intense scenes, but was always much smoother in the preview canvas.

This was strange to me, because the preview canvas is actually copied pixels from the output window. The fps counter was always reporting a steady 60fps.

My theory on why this was happening is this:
- Each screen has its own refresh rate, even if they are both 60hz, they are not in sync
- We were using `window.requestAnimationFrame` for rendering, which was always tied to the window of the UI
- Once sending an output, we were rendering onto the output window while tied to the wrong window's refresh rate, causing some mismatch in when the render happens and the screen refreshes
- This wasn't always causing a visible issue, but at times when there was some drift away from the perfect frame rate, the output was suffering a lot more than the preview, which was always synced

The solution:
- Our engine's renderer now has a `requestFrame` method, rather than `window.requestAnimationFrame` 
- This is always making use of `outputWindow.requestAnimationFrame`, which is set based on whether we're sending to a separate window or not

This means that now when sending output, you may see stuttering in the preview canvas, which is the right place for that to happen, not what the audience sees! Maybe later we want an option to disable copying of pixels to the preview or throttle it to 30fps or something.

Probably needs some[ follow up work](https://github.com/nudibranchrecords/hedron/issues/672) but I'm merging for now due to visible improvements in stuttering

### DIFF
--- a/apps/desktop/src/renderer/windows.ts
+++ b/apps/desktop/src/renderer/windows.ts
@@ -10,6 +10,7 @@ export const sendOutput = (display: Display): void => {
 
   outputWin.document.write('<div style="width:100vw;height:100vh;"></div>')
   outputWin.document.body.style.margin = '0'
+  outputWin.document.body.style.background = '#222'
   outputWin.document.body.style.cursor = 'none'
 
   const { x, y } = display.bounds
@@ -21,7 +22,7 @@ export const sendOutput = (display: Display): void => {
   })
 
   setTimeout(() => {
-    engine.setOutput(outputWin.document.querySelector('div') as HTMLDivElement)
+    engine.setOutput(outputWin.document.querySelector('div') as HTMLDivElement, outputWin)
   }, 1000)
 }
 
--- a/packages/engine/src/HedronEngine/HedronEngine.ts
+++ b/packages/engine/src/HedronEngine/HedronEngine.ts
@@ -551,8 +551,8 @@ export class HedronEngine {
     return this.renderer.createCanvas(containerEl)
   }
 
-  public setOutput(container: HTMLElement) {
-    this.renderer.setOutput(container)
+  public setOutput(container: HTMLElement, outputWindow: Window) {
+    this.renderer.setOutput(container, outputWindow)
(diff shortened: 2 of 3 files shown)


apps/desktop/src/renderer/windows.ts
```diff
@@ -21,7 +22,7 @@ export const sendOutput = (display: Display): void => {
   })
 
   setTimeout(() => {
-    engine.setOutput(outputWin.document.querySelector('div') as HTMLDivElement)
+    engine.setOutput(outputWin.document.querySelector('div') as HTMLDivElement, outputWin)
```
funwithtriangles, line 25: When setting up the output window, wenow  also send a reference of the window object

packages/engine/src/HedronEngine/HedronEngine.ts
```diff
@@ -706,7 +706,7 @@ export class HedronEngine {
       }
 
       if (this.paused) {
-        requestAnimationFrame(loop)
+        this.renderer.requestFrame(loop)
```
funwithtriangles, line 709: No longer using the native `requestAnimationFrame`, so that the refresh rate is always tied to the output window

packages/engine/src/world/Renderer.ts
```diff
@@ -161,10 +164,20 @@ export class Renderer {
     this.aspectRatio = ratio
   }
 
+  /**
+   * Requests the next animation frame, making sure to match the refresh rate of the output window
+   * @param callback The function to call on the next animation frame.
+   */
+  public requestFrame(callback: () => void): void {
+    this.rafId = this.outputWindow.requestAnimationFrame(callback)
+    this.frameCallback = callback
```
funwithtriangles, line 173: We need to hold on to `rafId` and `frameCallback` so we can cancel/restart the loop when needed

PR #675, opened 2026-07-23
get/set vector nodes using engine

funwithtriangles:
Fixes an issue where trying to use `engine.setParamValue` wasn't working when dealing with a vector type node (e.g. `rgb`). At the store level, we always want these nodes to be individual, but it makes sense for these methods on the engine to be clever and set the component nodes for vectors. Similarly, `engine.getParamValue` now returns an array of values if it is given the ID of a vector node.


packages/engine/src/store/actionCreators/updateParamValue.ts
```diff
@@ -8,13 +8,6 @@ export const createUpdateParamValue: SetterCreator<'updateParamValue'> = () => (
   tempParamValueBuffer[paramId] = value
 }
 
-export const createUpdateMultipleParamValues: SetterCreator<'updateMultipleParamValues'> =
-  () => (paramIds, values) => {
-    for (let i = 0; i < paramIds.length; i++) {
-      tempParamValueBuffer[paramIds[i]] = values[i]
-    }
-  }
-
```
funwithtriangles, line 17: Updating multiple values is now handled by the engine rather than the store

packages/ui-core/src/containers/NodeContainer/ParamColor/ParamColor.tsx
```diff
@@ -1,28 +1,19 @@
...
     ref.current?.updateColor(value as [number, number, number])
   })
 
-  const onVec3ValueChange = useOnParamVec3ValueChange(
-    vectorComponentIds[0],
-    vectorComponentIds[1],
-    vectorComponentIds[2],
-  )
-
-  return <ColorPicker ref={ref} onValueChange={onVec3ValueChange} />
```
funwithtriangles, line 27: was able to do lots of cleanup here and remove hooks we no longer need!

PR #680, opened 2026-07-29
make hedron more touchscreen friendly

cale-bradbury:
Makes hedron more touchscreen friendly
- Nodes can be tapped to display children/add inputs
- Wont grab slider handles when scrolling vertically
- Can grab the middle divider 

https://github.com/user-attachments/assets/687be007-4579-4035-a94b-0b4fd8ceb3ca


funwithtriangles:
Wonderful! I don't have a touchscreen to test with but now I want one 😅 thanks!

